PR #12, opened 2017-07-03
Custom value mapper

hannes-o-brunner:


hannes-o-brunner:
Introduced a bug which sbt assembly did not discover, should have done sbt clean assembly

PR #36, opened 2018-01-29
Improved parser performance for instance of, comparisons and filter expressions

ybasket:
Background: When I tried to run some small scripts (~10 lines) which involved heavy usage of contexts and functions, I noticed that parsing was awfully slow, it took up to 15 seconds to complete (while evaluation was pretty fast). Also, you seemed to have some trouble at Camunda with similar problems: https://blog.camunda.org/post/2016/08/dmn-performance-improvements/.

This commit eliminates some repeated parsing of common prefixes with optional suffix, like _instance of_ expressions where _expression7_ might already been parsed successfully, but would be thrown away if not followed by an _instance of_. The other changes cover similar cases. All tests still succeed, their execution time drops roughly from 10s to 2s on my machine (measured by one-time cold run, actual parsing benefit should be way higher as the tests include evaluation and parser initialization).

Would love to see these changes merged, if anything is unclear, just ask :)

### DIFF
--- a/feel-engine/src/main/scala/org/camunda/feel/parser/FeelParser.scala
+++ b/feel-engine/src/main/scala/org/camunda/feel/parser/FeelParser.scala
@@ -4,7 +4,6 @@ import org.camunda.feel._
 
 import scala.util.parsing.combinator.JavaTokenParsers
 import scala.util.Try
-import java.time.DateTimeException
 
 object FeelParser extends JavaTokenParsers {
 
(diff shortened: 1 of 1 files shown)


saig0:
Hi @ybasket , 

Thank you for your contribution. :+1: 

Your changes make the parser a bit more complicated but improve the speed a lot! 
I know that the current parser has some performance issues. At some point, I want to switch to another parsing framework, like https://github.com/lihaoyi/fastparse. 

Best regards,
Philipp

ybasket:
@saig0 Thanks for creating the project in the first place :)

Indeed, understanding the parser now takes a little more thinking, sadly, performance too often comes with a prize. The same goes for fastparse, I ported FEEL to it on Sunday for trying out fastparse (never used it before), but it looks uglier (fastparse doesn't support regex out of the box, some other combinators have to be defined on your own). Got all tests working in the end, but simplified/hacked some things (exact set of allowed characters for identifiers, some Unicode edge cases). If you want, I can push that branch later - maybe we can even make it proper ;)

Regards,
Yannick

saig0:
Thank's great! I would love to see what you archived. I also didn't work with FastParse before. :sweat_smile:

Issue #40, opened 2018-03-28
Date conversion function


richardmward:
Hi,

I am getting to grips with FEEL, and so I may be doing something wrong rather than this being an issue - so I apologise up front if that is the case!

If I try to run `date("2012-12-25") - date("2012-12-24") = duration("P1D")`, then I get a warning `WARN org.camunda.feel.FeelEngine - Suppressed failure: expected Number, Date, Time or Duration but found 'ValDate(2012-12-25)'` and returns `false` rather than `true`.

Interestingly if I change the expresssion to be `date and time("2012-12-25T00:00:00") - date and time("2012-12-24T00:00:00") = duration("P1D")`, then it does not cause a warning and returns `true` as expected.

I am executing the expressions via the `ScriptEngineManager` and am calling it using Java rather than Scala (which I have never experienced before this - and as such I have attempted to debug into it, but I have been unable work out how to!)

saig0:
Hi @richardmward ,

according to the spec, it seems that you can't subtract a date from a date. Please have a look at the chapter 10.3.2.12 (table 45, page 123).

Do you agree with me?
Or do I miss something?

Best regards,
Philipp

richardmward:
I had a feeling that it would be my limited knowledge of the FEEL specification - I agree with your interpretation that subtraction on a date is not a specified feature.

Many thanks for the reply - I'll close the issue.

saig0:
I understand your confusion and I would also expect such a function. Maybe we should just add the missing function.

saig0:
@richardmward , do you want to provide a pull request? 
I know that you have no experience in Scala but it's not that hard. You need to add some lines here: https://github.com/camunda/feel-scala/blob/master/feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala#L399   and write some test cases.

PR #41, opened 2018-03-28
support date subtract date and date subtract year month duration

richardmward:
Associated to #40 - My initial attempt supports `ValDate - ValDate` and `ValDate - ValYearMonthDuration`. I cannot figure out how to implement `ValDate - ValDayTimeDuration` (which would be required to do something like `date("2012-12-25") - duration("1D")`.

I also am slightly confused as to why the date subtraction produces negatives where I would expect positive and vice-versa. `time("10:00:00") - time("09:00:00")` returns `PT1H`, whereas `date("2012-12-25") - date("2012-12-24")` returns `P-1D`.

### DIFF
--- a/feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
+++ b/feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
@@ -383,6 +383,7 @@ class FeelInterpreter {
       case ValYearMonthDuration(y) => ValYearMonthDuration( x.plus(y).normalized )
       case ValLocalDateTime(y) => ValLocalDateTime( y.plus(x) )
       case ValDateTime(y) => ValDateTime( y.plus(x) )
+      case ValDate(y) => ValDate( y.plus(x) )
       case _ => error(y, s"expect Date-Time, or Year-Month-Duration but found '$x'")
     }
     case ValDayTimeDuration(x) => y match {
@@ -391,7 +392,12 @@ class FeelInterpreter {
       case ValDateTime(y) => ValDateTime( y.plus(x) )
       case ValLocalTime(y) => ValLocalTime( y.plus(x) )
       case ValTime(y) => ValTime( y.plus(x) )
+      case ValDate(y) => ValDate( y.atStartOfDay().plus(x).toLocalDate() )
       case _ => error(y, s"expect Date-Time, Time, or Day-Time-Duration but found '$x'")
+    }
+    case ValDate(x) => y match {
+      case ValDayTimeDuration(y) => ValDate( x.atStartOfDay().plus(y).toLocalDate() )
+      case ValYearMonthDuration(y) => ValDate( x.plus(y) )
     }
     case _ => error(x, s"expected Number, String, Date, Time or Duration but found '$x'")
   }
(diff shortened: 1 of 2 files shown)


richardmward:
I think there is probably more still wrong with that as I still can't execute my original expression `date("2012-12-25") - date("2012-12-24") = duration("P1D")` due to `Expect Year-Month-Duration but found 'ValDayTimeDuration(PT24H)'`

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
```diff
@@ -420,6 +420,11 @@ class FeelInterpreter {
       case ValDayTimeDuration(y) => ValDateTime( x.minus(y) )
       case _ => error(y, s"expect Time, or Year-Month-/Day-Time-Duration but found '$x'")
     }
+    case ValDate(x) => y match {
+      case ValDate(y) => ValYearMonthDuration( Period.between(x, y) )
```
saig0: The result should be "ValDayTimeDuration", similar to date-time. You must transform the date to a date-time with time = 0 and calculate the duration. Note that you need to swap the arguments in the `between` function.

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
```diff
@@ -420,6 +420,11 @@ class FeelInterpreter {
...
+      case ValYearMonthDuration(y) => ValDate( x.minus(y) )
```
saig0: It should also be possible to subtract a `ValDayTimeDuration`.

saig0:
Also, it should be possible to add a day-time- / year-month-duration to date (and the other way around).

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
richardmward: Of course - I need to turn the arguments around (I had copied my initial impl from ValTime in the subOp which just so happens to be the only subtraction that is `x, y` rather than `y, x`.

Could you expand on "transform the date to a date-time with time = 0" - is there an example of how that is done somewhere?

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
richardmward: Yes, agreed. I intended to do that, but I was stuck because of not knowing how to do the conversion - will add it in once I have got to grips with that.

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
saig0: Yep. Just call `date.atStartOfDay` which returns a date-time (thanks to Java 8 Date-Time API).

feel-engine/src/main/scala/org/camunda/feel/interpreter/FeelInterpreter.scala
richardmward: Hi @saig0, Should `ValDate - ValDayTimeDuration` return `ValDateTime` or `ValLocalDateTime`? What is the pattern for this?

saig0: The result should be ValDate :smile:  For example: 
`date("2012-12-25") - duration("P1D")` = `date("2012-12-24")` 
`date("2012-12-25") - duration("P1M")` = `date("2012-11-25")`

richardmward: I'm a little confused by this reply - if removing a ValDayTimeDuration, the the user may be removing a number of hours, not a complete day, thus the result would presumably need to be a date time, not just a date?

saig0: This is correct so far. But as a final step of the calculation, you have to transform the result back to a date - that means ignore the time part.

`date("2012-12-25") - duration("P1D")` = `date("2012-12-24")`
`date("2012-12-25") - duration("PT1H")` = `date("2012-12-24")`
`date("2012-12-25") - duration("PT1M")` = `date("2012-12-24")`
`date("2012-12-25") - duration("P1DT1H")` = `date("2012-12-23")`

Note that this feature will be part of DMN 1.2. I checked that the result should be a date.

richardmward:
Sorry for the delay (Easter!). I think this now does all the requested features.

saig0:
Great! Thank you very much for your contribution :tada:

PR #54, opened 2019-04-10
mention SPIN dependency

ingorichtsmeier:
As you configure everything with the pomin spring boot setups, mention the dependency to SPIN explicitly.


saig0:
Thank you :+1:

PR #78, opened 2020-01-20
CAM-11259: migrate camunda feel scala modules

koevskinikola:


saig0:
Instead of just removing the `example` module, we should think about migrating the examples and tests to the `engine` module.

tasso94:
Good point! So far we planned to move the examples to the Camunda BPM Examples repository but it actually makes sense to keep the examples in this repo and adjust them so that they only depend on `engine-feel`.

koevskinikola:
Hey @saig0, I brought back the `examples` in PR #81.

PR #79, opened 2020-01-20
CAM-11254: refactor namespace

koevskinikola:
:warning: Please don't merge before PR #78


saig0:
Just to clarify, why do you need to move every class into an `impl` package? 
As far as I understand, it is only public API if it is listed here: https://docs.camunda.org/manual/7.12/introduction/public-api/

So, if we don't add it there, for example, as the current FEEL engine, it is not considered as public API and can be changed at any time without taking care of the backward compatibility guarantees. 

However, some APIs like the FEEL engine and the `FunctionProvider` SPI should be stable anyways 
because they might be used by other applications.

tasso94:
I think we have two options here:
* We add the FEEL Engine to the list of public API modules; Next, we could discuss if we wanna move some "stable" interfaces out of the `impl` scope
* We don't add it to the list and keep the package namespace as it is

PR #83, opened 2020-01-28
chore(deps): update scala version to 2.13.1

tasso94:
related to CAM-11259

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 val shared = Seq(
   organization := "org.camunda.bpm.extension.feel.scala",
   version := "1.9.0-SNAPSHOT",
-  scalaVersion := "2.13.0",
+  scalaVersion := "2.13.1",
   resolvers += Resolver.mavenLocal,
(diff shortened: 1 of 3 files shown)


saig0:
BTW: can we enable Dependabot for this repo?

tasso94:
I've just enabled Depndabot. However, it seems like it does not work with sbt.

saig0:
Thank you for enabling the bot :+1: 

Yes. SBT is currently not supported (https://github.com/dependabot/dependabot-core/issues/352). But the primary build tool is Maven, so that is a good improvement :tada: 
Let's keep this topic in mind :)

tasso94:
@koevskinikola, it seems like `scala-parser-combinators_2.13` currently not compatible with Scala 2.13.1 (a warning is displayed). Let's don't merge this change for now.

koevskinikola:
@tasso94 I will close this PR until the `scala-parser-combinators_2.13` becomes compatible with Scala `2.13.1`, at which point we can re-open it.

PR #93, opened 2020-02-17
Add Camunda BPM integration module & compatibility tests

tasso94:
related to https://jira.camunda.com/browse/CAM-11289

### DIFF
--- a/feel-qa/camunda-bpm-platform-integration/pom.xml
+++ b/feel-qa/camunda-bpm-platform-integration/pom.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    
+    <artifactId>camunda-bpm-platform-integration</artifactId>
+    <version>1.11.0-SNAPSHOT</version>
+
+    <parent>
+        <groupId>org.camunda.feel</groupId>
+        <artifactId>feel-scala-root</artifactId>
+        <version>1.11.0-SNAPSHOT</version>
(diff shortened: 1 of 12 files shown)


saig0:
1) Why are you writing tests for the Camunda platform integration in the FEEL-Scala project?
It sounds like these tests should be written where the FEEL engine is used. For the integration in Zeebe, I'm writing tests in the Zeebe project and not here.

2) Why do you prefer `RefSpec` over `FlatSpec`? 
The other tests are written using the `FlatSpec`. It is ok to use a different test style. Please share your reason.

tasso94:
Hey Philipp,

* **Re 1:** we think that adding test cases to ensure the compatibility with Camunda BPM is a good idea; Like this breaking changes can immediately be spotted, discussed and resolved in a timely-manner
* **Re 2:** we haven't found a way to ignore test cases with `FlatSpec`; Could you give us some hints here?

Cheers,
Tassilo

saig0:
2) Using `FlatSpec`, a test can be ignored using the "tag" `ignore`: 

```scala
  it should "foo" in {
      ...
  }

  ignore should "bar" in {
    ...
  }
```

See http://www.scalatest.org/user_guide/tagging_your_tests

saig0:
1. I had a look at the tests and don't see any Camunda platform specifics. The tests use the normal API and nothing special. 
I guess that the Camunda platform integration should rely on the public API. So, we can also add these test cases to the other tests in the engine model.  
What makes the tests Camunda specific?

tasso94:
Hi Philipp,

thanks for sharing your thoughts. Here are our main reasons for separating the tests into a dedicated module:
* The tests are designed as blackbox tests only relying on the public API
  * We configured the value mapper via SPI exactly the way we use it in Camunda BPM; Configuring the `JavaValueMapper` right now via SPI in the main module would probably cause test failures
* Camunda BPM only relies on a subset of the `feel-scala` API; Not all API changes are meaningful for us and it makes sense to be able to distinguish if changes affect Camunda BPM or not

Cheers,
Tassilo

tasso94:
@koevskinikola, unassigned since I need to rework the PR

tasso94:
Close in favor of #104

PR #100, opened 2020-02-18
chore(project): define API guarantees

### LINKED ISSUE #95: Document API of the project

* This code base is owned by the Zeebe and the Camunda Runtime team for use in Zeebe as well as Camunda BPM
* We want to work on it independently, so that we do not have to sync between the teams for every change
* However, both teams need certain concepts to remain stable with new releases
* We have decided that we will use semantic versioning and agreed on a set of features that need to remain stable unless we build a major release
* If a team wants to make a breaking change, they must sync with the other team

* The API of this project is documented in the contributor's guide so that people are aware of what a breaking change would be

ThorbenLindhauer:
Describes what we consider API of this project and what changes must lead to a major release according to semantic versioning. Based on the discussion we had today.

@saig0, @tasso94: In this exceptional case, I think it would be good if one member of each team reviews that this captures everything as discussed.

Closes #95

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,7 @@ Scala code is formatted using [Scalafmt](https://scalameta.org/scalafmt/). The f
 
 Commit messages should include a short description of the changes and reference the issue.
 
-## API
+## Public API and Backwards Compatibility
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -71,3 +71,21 @@ Scala code is formatted using [Scalafmt](https://scalameta.org/scalafmt/). The f
 
 Commit messages should include a short description of the changes and reference the issue.
 
+## API
```
saig0:
```suggestion
## Public API and Backwards Compatibility
```

PR #116, opened 2020-03-06
feat(engine): disable external functions

### LINKED ISSUE #102: Question: Is it possible to prevent the use of `external` functions
Is it possible to disable the use of external functions so that users wouldn't have access to any java class in the JVM?

saig0:
Changes:

* external functions are disabled by default (security risk)
* new configuration option to enable external functions
* parsing of an expression fails if it contains an external function and the option is not enabled

Closes #102


docs/feel-expression.md
```diff
@@ -338,6 +338,8 @@ function(x,y) external {
 }
 ```
 
+:warning: External functions are disabled by default (security risk). Use the [FunctionProvider API](function-provider-spi) instead or enable external functions in the configuration (not recommended).
```
pihme: Unclear what "enable functions in the configuration" refers to. Maybe an example of how to do that would help

feel-engine/src/main/scala/org/camunda/feel/FeelEngine.scala
```diff
@@ -152,8 +167,19 @@ class FeelEngine(val functionProvider: FunctionProvider =
...
+
+    validator
+      .validateExpression(exp.expression)
+      .map(failure =>
+        Failure(
+          s"""validation of expression '${exp.text}' failed: ${failure.message}"""))
+      .toLeft(exp)
+  }
+
   def eval(exp: ParsedExpression, context: Context): EvalExpressionResult =
```
pihme: Question (based in insufficient Scala knowledge): Are these functions final, or is there a way to override these functions and skip the validation?

feel-engine/src/main/scala/org/camunda/feel/FeelEngine.scala
```diff
@@ -175,8 +201,10 @@ class FeelEngine(val functionProvider: FunctionProvider =
 
   def parseExpression(expression: String): Either[Failure, ParsedExpression] =
     parse(FeelParser.parseExpression, expression)
+      .flatMap(validate)
```
pihme: Question on design: Why do you validate here and not inside the parser?

feel-engine/src/main/scala/org/camunda/feel/impl/parser/ExpressionValidator.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.camunda.feel.syntaxtree.{Comparison, _}
+
+class ExpressionValidator(externalFunctionsEnabled: Boolean) {
+
+  def validateExpression(expression: Exp): Option[Failure] =
+    validate(expression).headOption
+
+  private def validate(exp: Exp): List[Failure] = exp match {
+    // validate expression
+    case JavaFunctionInvocation(_, _, _) if !externalFunctionsEnabled =>
```
pihme: Why just JavaFunctionInvocation? Or is this the only case for external function invocations in this code base?

feel-engine/src/main/scala/org/camunda/feel/impl/parser/ExpressionValidator.scala
```diff
@@ -0,0 +1,83 @@
...
+      List(Failure(
+        "External functions are disabled. Use the FunctionProvider SPI (recommended) or enable external function in the configuration."))
+
+    // delegate to inner expression
+    case ConstList(items) => items.flatMap(validate)
```
pihme: This list seems very verbose and error prone. What happens if you extend the expressions? Will the compiler detect that a case is not covered here? (not sure if the last line here is something like a default case)
I would rather expect something like a visitor pattern here. Not sure what the equivalent is in a FP style

feel-engine/src/test/scala/org/camunda/feel/ExternalFunctionsTest.scala
```diff
@@ -0,0 +1,82 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.camunda.feel
+
+import org.camunda.feel.FeelEngine.{Configuration, Failure}
+import org.camunda.feel.syntaxtree.ParsedExpression
+import org.scalatest.{FlatSpec, Matchers}
+
+class ExternalFunctionsTest extends FlatSpec with Matchers {
```
pihme: Not sure about the test scope and the name of this test class. If this test is only to test the switch, then the content is Ok, but the name promises maybe a little too much.

If this test is for "everything related to external functions", I would also expect test cases and/or assertions where the result of an external function parsing and evaluation is verified.

Also I would expect a small unit test for the ExpressionValidator

docs/feel-expression.md
pihme: PS: became clear after doing the whole review. But just reading the docs it left me puzzled and misdirected me into looking for a configuration file.

saig0: Agree. The configuration (or how to use/setup the engine) is not well documented yet.

feel-engine/src/main/scala/org/camunda/feel/FeelEngine.scala
saig0: As a user of the FEEL engine, there is no way to skip the validation.

feel-engine/src/main/scala/org/camunda/feel/FeelEngine.scala
saig0: I tried to put it inside the parser but that does not work with the parser library (in a good way). So, it decided to validate after the parsing.

feel-engine/src/main/scala/org/camunda/feel/impl/parser/ExpressionValidator.scala
saig0: Yes. External functions are parsed as `JavaFunctionInvocation`. (https://github.com/camunda/feel-scala/blob/master/feel-engine/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala#L430-L434)

feel-engine/src/main/scala/org/camunda/feel/impl/parser/ExpressionValidator.scala
saig0: Yes, the compiler can see if all cases are covered because `Exp` is a sealed trait. If we introduce a new `Exp` type then it needs to be added also to the validation. 

`case _ => ` is the default case

Yes, this works like a visitor pattern.

feel-engine/src/test/scala/org/camunda/feel/ExternalFunctionsTest.scala
saig0: Do you have a suggestion for a name? This class should only test that external functions can be enabled and disabled.

saig0: A unit test for the `ExpressionValidator` itself would be a bit clunky to write because it works on the parsed expression. I think it is more stable to use the FEEL syntax directly instead of relying on the syntax tree. 
However, I could cover more cases. I just picked the most obvious case with the context. I think all other cases would also use the context as part of the expression.

pihme: Something like `ExternalFunctionsFlag/Switch/SettingTest`

saig0:
@pihme, thank you for the review. I renamed the test class and described how to enable the external function in the README. Please have a look.

feel-engine/README.md
```diff
@@ -67,6 +67,14 @@ Or using Java:
...
 ```
 
+Enable FEEL External-Functions in the configuration: (not recommended)
+
+```
+new FeelEngine(configuration = Configuration(externalFunctionsEnabled = true))
+// or
+new FeelEngine.Builder().enableExternalFunctions(true).build()
+```
+
```
pihme:
```suggestion
Enabling external functions is not recommended because external functions make it possible to call arbitrary Java methods. This poses a potential security risk. 

Enabling this feature can be safe if you have full control of the expressions that will be evaluated (i.e. none of them are entered by a user or generated by an external system). In all other cases this feature is considered unsafe.

An alternative to external functions is the use of (Function Providers)[https://camunda.github.io/feel-scala/function-provider-spi]
```

PR #117, opened 2020-03-09
Update documentation (expression/unary-tests)

saig0:
* correct the boolean disjunction/conjunction logic for null values (expression section)
* describe when a unary-tests expression is true (unary-tests section)

References:
* DMN 1.2 specification table 45 page 128
* DMN 1.2 specification section 8.3.3 page 90

### DIFF
--- a/docs/feel-expression.md
+++ b/docs/feel-expression.md
@@ -158,7 +158,7 @@ true and false
 // false
 
 true and null        
-// false
+// null
 
 false and null
 // false
@@ -175,7 +175,7 @@ true or null
 // true
 
 false or null  
-// false
+// null
 ```
 
 ### If Expression
--- a/docs/feel-unary-tests.md
+++ b/docs/feel-unary-tests.md
@@ -8,6 +8,11 @@ Unary-Tests can be used only for input entries of a decision table. They are a s
 
 The result of the expression must be either `true` or `false`.
 
+An unary-tests expression is `true` if one of the following conditions is fulfilled:
+* the expression evaluates to `true` when the input value is applied to it
+* the expression evaluates to a list and the input value is equal to at least one of the values in that list
+* the expression evaluates to a value and the input value is equal to that value 
+
 ### Comparison
 
 Compare the input value to `x`.
@@ -28,11 +33,11 @@ Compare the input value to `x`.
   * year-month-duration
   * day-time-duration 
 
-### Inveral
+### Interval
 
 Test if the input value is within the interval `x` and `y`.
 
-An interval can be open `(x..y)` / `]x..y[` or close `[x..y]`. If the interval is open when the value is not included.
+An interval can be open `(x..y)` / `]x..y[` or closed `[x..y]`. If the interval is open then the value is not included.
 
 ```js
 (2..5)


docs/feel-expression.md
```diff
@@ -158,7 +158,7 @@ true and false
 // false
 
 true and null        
-// false
+// null
```
pihme: Did this change because of your change? Or was this wrong in the documentation?

docs/feel-expression.md
```diff
@@ -175,7 +175,7 @@ true or null
 // true
 
 false or null  
-// false
```
pihme: Did this change because of your change? Or was this wrong in the documentation?

pihme: Apparently I can only comment near to lines where you made changes.

This comment is about places where you didn't make changes:

- "Evaluate a Unary Tests" -> "Evaluate a Unary Tests" / "Evalaute Unary Tests"
- Below is a section on external functions. This section would benefit from a hint that external functions are disabled by default

docs/feel-unary-tests.md
```diff
@@ -6,7 +6,10 @@ title: FEEL Language Reference: Unary Tests
 
 Unary-Tests can be used only for input entries of a decision table. They are a special kind of expression with additional operators. The operators get the value of the input expression implicitly as the first argument. 
 
-The result of the expression must be either `true` or `false`.
```
pihme: I would keep this sentence in. Otherwise you only describe under what condition the expression evaluates to true, but you don't say what the other result is.

docs/feel-unary-tests.md
```diff
@@ -28,7 +31,7 @@ Compare the input value to `x`.
   * year-month-duration
   * day-time-duration 
 
-### Inveral
+### Interval
 
 Test if the input value is within the interval `x` and `y`.
```
pihme: Next line: 
An interval can be open or close**d**. If the interval is open **then** the value _at the boundary_ is not included.

docs/feel-expression.md
saig0: Nope. The documentation was a lie :sweat_smile:

docs/feel-expression.md
saig0: Again. The documentation was a lie.

I added the comment in the other PR: https://github.com/camunda/feel-scala/pull/116/files#diff-eea0eae9141e7419ee431be85fc00127R341

docs/feel-unary-tests.md
saig0: Ok. I'll leave it :)

docs/feel-unary-tests.md
saig0: Good catch!

pihme:
green light from my side

saig0:
@pihme, thank you for the review. I applied your hints. Please have a look.

pihme:
You still did not address my heading comment (like I said, I might have pressed the wrong buttons in Github):
"I would also like to see a page/example on how to use the engine in your own code as part of the documentation. After all, that is what the headline on the front page promises to teach me."

Basically, the documentation promises me to teach me how to use the engine, but then all examples are about FEEL expressions. What is missing is an example of how to instantiate the engine. 

You do have an example at https://github.com/camunda/feel-scala/tree/master/feel-engine#usage but I haven't found an example in the documentation itself. Also this example would benefit of a section explaining the new flag.

This comment should not prevent a merge. Just a feedback for you to consider.

saig0:
I appreciate your comment and agree with it. I created a new issue (#119) to fill this gap in the documentation.

pihme:
Review approved

(Do I need to start bors or press any buttons?)

PR #121, opened 2020-04-15
feat(ci): build maven project with GitHub Actions

lwille:
related to INFRA-1253

we picked this project to be the first to start building using GitHub Actions.

* a Maven build is triggered for every push :heavy_check_mark: 
* a SNAPSHOT deploy is triggered for every push to master :heavy_check_mark: 
* when a release is published, a Maven release with the appropriate tag is deployed and the development version is increased :heavy_check_mark: 
* a master build is scheduled every day (without SNAPSHOT release) :heavy_check_mark:


AndreaGiardini:
The old Jenkinsfile had:
```
buildMavenAndDeployToMavenCentral([jdk:8, mvn:3.5, licenseCheck:true, publishZipArtifactToCamundaOrg:true])
```

So I think our maven target should also include `com.mycila:license-maven-plugin:check`.
@lwille what do you think?

lwille:
@saig0 we're now supporting attaching Release Assets like that:
image

that's what it'll look like when attached to Release (taken from our demo repo):
image

It's also possible to attach single files too (without zipping), but it'll require duplicating the `upload-release-asset` step for every file - that's why we chose to zip everything into one file.

PR #122, opened 2020-04-22
INFRA-1153: build maven project with GitHub Actions 

lwille:
- reset base to master before merging
- contains all changes from and closes #121


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
+name: Build and publish
```
AndreaGiardini, line 1: Let's have a more meaningful name and eventually write down a couple of comments to better explain the workflow

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+          echo "::set-output name=type::RELEASE"
+        fi
+      id: release-detection
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    - name: Import Secrets
+      uses: RichiCoder1/vault-action@v0.10.2
+      with:
+        url: ${{ secrets.STAGE_VAULT_ADDR }}
```
AndreaGiardini, line 26: I guess that this shouldn't be stage but production
I recommend using `TODO` in your code whenever you add a change that you want to roll back later

AndreaGiardini: Same applies to the next 3 lines

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: Update version in pom.xml (Release only)
+      if: github.event.release
+      run: mvn -B versions:set -DnewVersion=${{ github.event.release.tag_name }} -DgenerateBackupPoms=false
+    - name: Run Maven (${{ steps.release-detection.outputs.type }})
+      run: mvn -B clean install
+    - name: Publish package (${{ steps.release-detection.outputs.type }})
+      run: mvn -B deploy
+    - name: Artchive artifacts
```
AndreaGiardini, line 54: Artchive -> Archive

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+      run: |-
+        ZIPFILE=${{github.event.repository.name}}-${{github.event.release.tag_name}}.zip
+        zip $ZIPFILE $(find target/nexus-staging -name "*.jar")
+        echo "::set-output name=filename::${ZIPFILE}"
+      id: zip
+    - name: Publish artifacts (Release only)
```
AndreaGiardini, line 60: I would replace it with `Publish artifacts to Github Releases` otherwise people might get confused with the `Publish package` step (maybe also in that case it makes sense to have a longer name)

AndreaGiardini:
I left some comments that need to be addressed before merge

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,79 @@
...
+        server-password: NEXUS_APP_CAMUNDA_COM_PSW
+    - name: Cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.m2/repository
+        key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
+        restore-keys: |
+          ${{ runner.os }}-maven-
+    - name: Update version in pom.xml (Release only)
+      if: github.event.release
```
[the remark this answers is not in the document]
lwille: good suggestion!

.github/workflows/build.yml
```diff
@@ -1,5 +1,5 @@
-name: Build
-on:
+name: Build project with Maven
+on:                                          # yamllint disable-line rule:truthy
```
AndreaGiardini, line 2: Since `yamllint` is not included in the repo I would not include this comment

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,85 @@
+# If this workflow is triggered by a push to master, it
+#        deploys a SNAPSHOT
+# If this workflow is triggered by publishing a Release, it
+#        deploys a RELEASE with the selected version
+#        updates the project version by incrementing the patch version
+#        commits the version update change to the repository's default branch.
+name: Deploy artifacts with Maven
+on:                                          # yamllint disable-line rule:truthy
```
AndreaGiardini, line 8: Since `yamllint` is not included in the repo I would not include this comment

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,85 @@
...
+    - name: Import Secrets
+      uses: RichiCoder1/vault-action@v0.10.2
+      with:
+        url: ${{ secrets.VAULT_ADDR }}
+        method: approle
+        roleId: ${{ secrets.VAULT_ROLE_ID }}
+        secretId: ${{ secrets.VAULT_SECRET_ID }}
+        secrets: |
+          common/jenkins/main NEXUS_APP_CAMUNDA_COM_PSW | NEXUS_APP_CAMUNDA_COM_PSW;
+          common/jenkins/main NEXUS_APP_CAMUNDA_COM_USR | NEXUS_APP_CAMUNDA_COM_USR;
```
AndreaGiardini, line 38: Is this account the common `jenkins` shared account we are trying to migrate out of? Or it's a new account that we use for github?

AndreaGiardini:
A couple of more comments and then i would say it's ready to go

.github/workflows/deploy.yml
lwille: that's `ci-cambpm`. I wanted to avoid having to ask IT to create an account.

However, as a preparation we can create a dedicated entry for GitHub Actions.

AndreaGiardini:
:+1: LGTM

PR #135, opened 2020-06-05
chore(project): morph into a single module project

saig0:
* moved the feel-engine module into root
* remove example module - examples and tests are covered by the engine module
* remove SBT as build tool in favor of Maven - maintaining both build tools and keep them in sync is too much work


saig0:
@koevskinikola, done :tada:  Thanks for checking the compatibility :+1:

PR #142, opened 2020-06-10
fix(engine): compare context and list values

### LINKED ISSUE #134: A context can not be compared to another context
**Describe the bug**
* I can not check if one context is equal to another context 

**To Reproduce**
* evaluate the expression `{} = {}`
* the evaluation fails with the message:
`ValError(expected Number, Boolean, String, Date, Time or Duration but found 'ValContext(org.camunda.feel.impl.interpreter.EvalContext@5b7a5baa)')`

**Expected behavior**
* a context can be compared to another context
* the context is equal to another context if all entries are equal

**Environment**
Feel Engine 1.11.0

Related issue: https://jira.camunda.com/browse/CAM-11319

saig0:
* in an expression, a list/context value can be compared to another list/context value
  * two contexts are equal if they have the same keys and every entry is equal
  * two lists are equal if they have the same size and every item is equal
  * ignore failures when comparing the entries/items (e.g. different types)
* in a unary-test, a list/context input value can be compared to a list/context value

closes #134

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -21,7 +21,6 @@ import java.time.{Duration, Period}
 import org.camunda.feel.FeelEngine.UnaryTests
 import org.camunda.feel._
 import org.camunda.feel.context.Context
-import org.camunda.feel.impl.parser._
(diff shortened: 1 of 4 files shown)


tasso94:
**Please Note** 

The commit references the wrong CAM issue. The correct one is: https://jira.camunda.com/browse/CAM-12065

PR #165, opened 2020-07-18
Extends decimal() built-in function by adding the parameter for rounding mode

### LINKED ISSUE #167: Add built-in function for (commercial) rounding up
When performing DMN calculations, it is very often necessary to round off numbers commercially. (= `RoundingMode.HALF_UP`)

* **Is the function round() part of the DMN/FEEL specification?**

No, it is not. Source: https://www.omg.org/spec/DMN/1.2/PDF  (Page 149)

* **how is the function different from decimal(), floor() and ceiling()?**

The round()-function uses `RoundingMode.HALF_UP`  (decimal() uses `HALF_EVEN`, floor() uses `FLOOR`, ceiling() uses `CEILING`

* **in which context do you need this function? (i.e. Camunda BPM platform, Zeebe, something else)**

This function is needed in context of Camunda BPM platform (DMN calculations), but can be probably used in Zeebe too.

Referenced Pull-Request: #165

bigpuritz:
## Description

 
* adds new built-in function for rounding numbers (`RoundingMode.HALF_UP`)

closes #167


saig0:
Hi @bigpuritz,

thank you for your contribution :tada: 

Before merging your PR, please create an issue. An issue helps to track the change and help me to understand what is the problem you want to solve with the PR.

The following questions should be answered:
* is the function `round()` part of the DMN/FEEL specification?
* how is the function different from `decimal()`, `floor()` and `ceiling()`?
* in which context do you need this function? (i.e. Camunda BPM platform, Zeebe, something else)

src/main/scala/org/camunda/feel/impl/interpreter/BuiltinFunctions.scala
```diff
@@ -1034,6 +1033,13 @@ object BuiltinFunctions extends FunctionProvider {
       case e => error(e)
     })
 
+  def decimalFunction3 =
+    ValFunction(List("n", "scale", "mode"), _ match {
+      case List(ValNumber(n), ValNumber(scale), ValString(mode)) =>
+        ValNumber(n.setScale(scale.intValue, RoundingMode.withName(mode)))
```
saig0, line 1039: To avoid exceptions, we should check if the mode is a valid `RoundingMode` (in a case-insensitive way). If not then we should return a `ValError` that also lists all valid rounding modes.

src/test/scala/org/camunda/feel/interpreter/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -39,6 +39,17 @@ class BuiltinNumberFunctionsTest
     eval(" decimal(2.5, 0) ") should be(ValNumber(2))
   }
 
+  "A decimal() function with rounding mode" should "return number with a given scale" in {
```
saig0, line 42: We can group all tests for one function together.

```suggestion
  it should "round the number using the given rounding mode" in {
```

src/test/scala/org/camunda/feel/interpreter/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -39,6 +39,17 @@ class BuiltinNumberFunctionsTest
...
+
+    eval(""" decimal(1.5, 0, "HALF_EVEN") """) should be(ValNumber(2))
+    eval(""" decimal(1.5, 0, "HALF_UP") """) should be(ValNumber(2))
+    eval(""" decimal(1.5, 0, "UP") """) should be(ValNumber(2))
+    eval(""" decimal(1.5, 0, "HALF_DOWN") """) should be(ValNumber(1))
+    eval(""" decimal(1.5, 0, "DOWN") """) should be(ValNumber(1))
+    eval(""" decimal(1.5, 0, "FLOOR") """) should be(ValNumber(1))
+    eval(""" decimal(1.5, 0, "CEILING") """) should be(ValNumber(2))
```
saig0, line 50: We should provide more examples to see the differences in the rounding modes.

Additionally, we should add a test case if the given mode is not valid.

Issue #91, opened 2020-02-12
Public API must contain all classes/objects/traits which are exposed in other public APIs


menski:
Currently the Public API exposes classes/objects/traits from `impl` packages, for example `Context` https://github.com/camunda/feel-scala/blob/9760122317f34e184503ef9b534bda5a211fcff5/feel-engine/src/main/scala/org/camunda/feel/FeelEngine.scala#L97

This should be aligned so that no `impl` classes/objects/traits are exposed.

tasso94:
Fixed with #94

PR #170, opened 2020-07-31
Enable Clirr for public API compatibility check

saig0:
## Description

* using the Clirr maven plugin to verify that we don't break the public API (i.e. it is ABI-compatible with a previous version)

## Related issues

related to #91


pom.xml
```diff
@@ -206,6 +208,27 @@
...
+        <artifactId>clirr-maven-plugin</artifactId>
+        <version>2.8</version>
+        <executions>
+          <execution>
+            <goals>
+              <goal>check</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
```
koevskinikola, line 222: I think an "old" version for comparison needs to be defined here (the `feel.version.old` property needs to be defined inside the properties tag):
```suggestion
        <configuration>
          <comparisonVersion>${feel.version.old}</comparisonVersion>
```

saig0:
@koevskinikola , I'm not sure if this is true. In the documentation (https://www.mojohaus.org/clirr-maven-plugin/examples/specific-version.html), it says that:

> By default, the Clirr Maven Plugin compares the current code against the latest released version, which is lower than the current version.

So, it should work in the following way:
* release `1.12.0` -> compare with `1.11.2`
* release `1.11.3` -> compare with `1.11.2`
* release `1.12.1` -> compare with `1.12.0`

Okay, I'm sure about the last point. But we can see and add the version tag if we run into the problem :sweat_smile:

koevskinikola:
That makes sense. Looks good then. :)

PR #176, opened 2020-08-25
Compare not existing variables with null

### LINKED ISSUE #150: Compare not existing variables with Null
**Describe the bug**
As a user, I want to check if a variable exists and is not `null`.

For example:
```
x != null
```

This expression returns `false` if the variable exists and is `null`
But it returns `true` if the variable doesn't exist. 

This behavior is counter-intuitive and may confuse most users.  

**To Reproduce**
Expression: `x != null`

1. evaluate the expression with variables: `x` = `null` -> `false`
1. evaluate the expression without variables -> `true`

**Expected behavior**
A non-existing variable is handled as `null` value. It can be compared with `null` and returns `true`.

In some cases, it can be useful to have an option to check if a variable exists or not. For example, by having a built-in function `exist(string): boolean`.

**Environment**
* FEEL engine version: 1.11.2
* Affects: 
  * Zeebe broker: [0.23.0, 0.24.0] https://github.com/zeebe-io/zeebe/issues/4212

saig0:
## Description

* comparing a non-existing variable with `null` returns `true`
* comparing a property with `null` returns `true` if the property is `null`, if it doesn't exist, or if its variable is `null` or doesn't exist 
* new built-in function `is defined()` that returns `true` if the given value is present (i.e. is not an error)

## Related issues

closes #150

### DIFF
--- a/docs/develop/feel-built-in-functions.md
+++ b/docs/develop/feel-built-in-functions.md
@@ -135,6 +135,30 @@ not(true)
 // false
 ```
 
+### is defined()
+
+Checks if a given value is defined or not. A value is defined if it exists, and it is an instance of one of the FEEL data types including `null`.
+
+The function can be used to check if a variable, or a context entry (e.g. a property of a variable) exists. It allows differentiating between a variable that is `null` and a value that doesn't exist.   
+
+* parameters:
+  * `value`: any
+* result: boolean
+
+```js
+is defined(1)
+// true
+
+is defined(null)
+// true
+
+is defined(x)
+// false - if no variable "x" exists
+
+is defined(x.y)
+// false - if no variable "x" exists or it doesn't have a property "y"
+```
+
 ## String Functions 
 
 ### substring()
--- a/docs/develop/feel-expression.md
+++ b/docs/develop/feel-expression.md
@@ -138,13 +138,30 @@ duration("P1Y") / duration("P1M")
 | greater than or equal | `>=` | `>= 10` |
 | between | `between _ and _` | `x between 3 and 9` |
 
-* less than/greater than/between are only supported for: 
+The operators less than, greater than, and between are only supported for: 
   * number
   * date
   * time
   * date-time
   * year-month-duration
   * day-time-duration 
+  
+Any value can be compared with `null` to check if it is equal to `null`, or if it exists. Comparing `null` to a value different from `null` results in `false`. It returns `true` if the value, or the context entry (e.g. the property of a variable) is `null` or doesn't exist. The built-in function [is defined()](feel-built-in-functions.md#is-defined) can be used to differentiate between a value that is `null` and a value that doesn't exist. 
+
+```js
+null = null
+// true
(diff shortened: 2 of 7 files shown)


docs/develop/feel-built-in-functions.md
```diff
@@ -135,6 +135,28 @@ not(true)
 // false
 ```
 
+### is defined()
+
+Checks if a given value is defined or not. Can be used to check if a variable, or a property of an object exists.
```
pihme, line 140: The explanation here could be a little more detailed; especially the subtle difference between checking for null and existence/non-existence. Also the examples below could explicitly state that it will return true if the variable exists.

docs/develop/feel-expression.md
```diff
@@ -138,13 +138,29 @@ duration("P1Y") / duration("P1M")
...
-* less than/greater than/between are only supported for: 
+The operators less than, greater than, and between are only supported for: 
   * number
   * date
   * time
   * date-time
   * year-month-duration
   * day-time-duration 
+  
+Any value can be compared with `null` to check if it is equal to `null`, or if it exists. Comparing `null` to a value different from `null` results in `false`. It returns `true` if the value, or the property is `null` or doesn't exist.
```
pihme, line 149: Maybe add a reference to "is defined()" here and explain that this is a more precise check for existence

src/main/scala/org/camunda/feel/impl/interpreter/BuiltinFunctions.scala
```diff
@@ -1150,6 +1153,15 @@ object BuiltinFunctions extends FunctionProvider {
       }
     )
 
+  private def isDefinedFunction =
```
pihme, line 1156: (Optional) maybe replace it with "is undefined"

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
```diff
@@ -75,4 +75,19 @@ class InterpreterExpressionTest
     eval("null != null") should be(ValBoolean(false))
   }
 
+  it should "compare to not existing variable" in {
```
pihme, line 78: Please add test cases for existing variables which are null

```eval("null = {"x": null}.x) should be(ValBoolean(true))```

Same for property

pihme (approved):
Looking forward to using this.

Minor suggestions for changes.

src/main/scala/org/camunda/feel/impl/interpreter/BuiltinFunctions.scala
saig0: I thought about it. I prefer `is defined(x)` because it looks more intuitive when inverting it with `not(is defined(x))`. Instead of having a double negation `not(is undefined(x))`.

saig0:
@pihme thank you for the fast review :+1: 
I applied most of the hints :)

PR #185, opened 2020-09-16
Allow variable names with date-time keywords

### LINKED ISSUE #184: Evaluation fails if the variable name contains a date-time keyword
**Describe the bug**

The evaluation of an expression fails if a variable name contains a date-time keyword (e.g. `date`, `time`, `duration`). 

```
failed to evaluate expression 'duration(timer_duration)': no variable found for name 'cellInput'
```

Related to #147

**To Reproduce**
1. Evaluate the expression `timer_duration` with a variable "timer_duration" 
2. Verify that the evaluation failed with a message `no variable found for name 'cellInput'`

**Expected behavior**
A variable name can contain a date-time keyword.

**Environment**
* FEEL engine version: 1.12.0
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

saig0:
## Description

* adjust the parser to allow variable names with a date-time keyword 

## Related issues

closes #184

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -44,11 +44,6 @@ object FeelParser extends JavaTokenParsers {
     | "between\\b".r
     | "instance\\b".r | "of\\b".r)
 
-  private lazy val dateTimeWords: Parser[String] = ("date and time".r
-    | "date".r
-    | "time".r
-    | "duration".r)
-
   // list of built-in function names with whitespaces
   // -- other names match the 'function name' pattern
   private lazy val builtinFunctionName: Parser[String] = ("date and time"
@@ -118,7 +113,7 @@ object FeelParser extends JavaTokenParsers {
   private lazy val expression8 = functionInvocation | builtinFunctionInvocation | filteredExpression9
   // 2 i)
   private lazy val expression9 =
-    not(dateTimeWords) ~> name ^^ (n => Ref(List(n))) |
+    not(dateTimeLiteral) ~> name ^^ (n => Ref(List(n))) |
       literal |
       "?" ^^^ ConstInputValue |
       simplePositiveUnaryTest |
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
@@ -144,7 +144,10 @@ class InterpreterExpressionTest
     "elseY",
     "betweenXandY",
     "notThis",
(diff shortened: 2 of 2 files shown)


src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -118,7 +113,7 @@ object FeelParser extends JavaTokenParsers {
   private lazy val expression8 = functionInvocation | builtinFunctionInvocation | filteredExpression9
   // 2 i)
   private lazy val expression9 =
-    not(dateTimeWords) ~> name ^^ (n => Ref(List(n))) |
+    not(dateTimeLiteral) ~> name ^^ (n => Ref(List(n))) |
```
korthout, line 116: `dateTimeLiteral` seems to be defined for the date-time literals OR it is a failure. 

```scala
// Line 211 of this file
private lazy val dateTimeLiteral: Parser[Exp] =
    "date" ~ "(" ~> stringLiteralWithQuotes <~ ")" ^^ parseDate |
      "time" ~ "(" ~> stringLiteralWithQuotes <~ ")" ^^ parseTime |
      "date and time" ~ "(" ~> stringLiteralWithQuotes <~ ")" ^^ parseDateTime |
      "duration" ~ "(" ~> stringLiteralWithQuotes <~ ")" ^^ parseDuration |
      failure("expected date time literal")
```

Wouldn't that cause a problem here?

korthout (commented):
Test looks good, but I have a question about the implementation.

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
saig0: Nope. All tests pass :wink: 

The failure is defined in the grammar to produce a nicer failure message. It would be returned if the parser can not backtrack from the `dateTimeLiteral` parser (i.e. if there are no other parsers that match the token).

Does this make sense for you?

korthout: Ah okay. But why then does this work fine? Is it only because of backtracking? expression9 is used somewhere and the parser can still backtrack to a next part of the grammar definition; or is this `not()` method able to deal with a failure as a `not(false)`? Also does the failure method return `Parser[Exp]` type?

saig0: Yes. Yes. And no :sweat_smile: 

Let's have a quick chat :wink:

korthout: Awesome! Then LGTM. Have a 🌮

korthout (approved):
LGTM 🌮

PR #193, opened 2020-09-29
fix: fixed oddFunction to eval negative numbers

### LINKED ISSUE #151: odd() function doesn't work for negative numbers
**Describe the bug**
The `odd()` function produces a wrong result for negative numbers.

**To Reproduce**
Expression: `odd(-1)` returns `false` but should be `true`.

**Expected behavior**
`odd(-1)` returns `true`

**Environment**
* FEEL engine version: 1.11.2

kaaquist:
Signed-off-by: Kasper Aaquist Johansen <kasperaaquist@gmail.com>

## Description

* Made oddFunction return `true` when evaluating negative numbers.  

closes #151

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
@@ -5,6 +5,7 @@ import org.camunda.feel.impl.builtin.BuiltinFunction.builtinFunction
 import org.camunda.feel.syntaxtree._
 
 import scala.math.BigDecimal.RoundingMode
+import scala.math.Numeric.BigDecimalAsIfIntegral.abs
 
 object NumericBuiltinFunctions {
 
@@ -100,7 +101,7 @@ object NumericBuiltinFunctions {
 
   private def oddFunction =
     builtinFunction(params = List("number"), invoke = {
-      case List(ValNumber(n)) => ValBoolean(n % 2 == 1)
+      case List(ValNumber(n)) => ValBoolean(abs(n) % 2 == 1)
     })
 
   private def evenFunction =
--- a/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
@@ -5,7 +5,6 @@ import org.camunda.feel.impl.builtin.BuiltinFunction.builtinFunction
 import org.camunda.feel.syntaxtree._
 
 import scala.math.BigDecimal.RoundingMode
-import scala.math.Numeric.BigDecimalAsIfIntegral.abs
 
 object NumericBuiltinFunctions {
 
@@ -101,7 +100,7 @@ object NumericBuiltinFunctions {
 
   private def oddFunction =
     builtinFunction(params = List("number"), invoke = {
-      case List(ValNumber(n)) => ValBoolean(abs(n) % 2 == 1)
+      case List(ValNumber(n)) => ValBoolean(n.abs % 2 == 1)
     })
 
   private def evenFunction =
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
```diff
@@ -100,7 +101,7 @@ object NumericBuiltinFunctions {
 
   private def oddFunction =
     builtinFunction(params = List("number"), invoke = {
-      case List(ValNumber(n)) => ValBoolean(n % 2 == 1)
+      case List(ValNumber(n)) => ValBoolean(abs(n) % 2 == 1)
```
saig0, line 104: We can use the function `abs` directly on the `BigDecimal`. 

```suggestion
      case List(ValNumber(n)) => ValBoolean(n.abs % 2 == 1)
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -205,6 +205,8 @@ class BuiltinNumberFunctionsTest
 
     eval(" odd(5) ") should be(ValBoolean(true))
     eval(" odd(2) ") should be(ValBoolean(false))
+    eval(" odd(-5)") should be(ValBoolean(true))
+    eval(" odd(-2)") should be(ValBoolean(false))
```
saig0, line 209: In order to separate the cases, we can create a new test function below. For example:

``` 
it should "return true if a negative number is odd" in {
    eval(" odd(-5)") should be(ValBoolean(true))
    eval(" odd(-2)") should be(ValBoolean(false))
}
```

saig0 (changes requested):
@kaaquist thank you for your contribution :cookie: 

Looks good :+1: Just two comments for improvements :)

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
kaaquist: Separated code into its own test.

src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
kaaquist: Done. I did actually see that you also had a similar usages of `.abs` in the absFunction, so make sense.

src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
```diff
@@ -5,6 +5,7 @@ import org.camunda.feel.impl.builtin.BuiltinFunction.builtinFunction
 import org.camunda.feel.syntaxtree._
 
 import scala.math.BigDecimal.RoundingMode
+import scala.math.Numeric.BigDecimalAsIfIntegral.abs
```
saig0, line 8: We can remove the unused import :)

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -207,6 +207,12 @@ class BuiltinNumberFunctionsTest
     eval(" odd(2) ") should be(ValBoolean(false))
   }
 
+  "A odd() function" should "return true if negative number is odd" in {
```
saig0, line 210: Here, we can use the `it` function to combine both cases under one root :)

```suggestion
  it should "return true if negative number is odd" in {
```

saig0 (changes requested):
Two tiny points left :candy:

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
kaaquist: done.

saig0 (approved):
Thanks @kaaquist :tada:

PR #194, opened 2020-09-30
feat: added weekday for date and datetime, issue #160

### LINKED ISSUE #160: I can access the weekday as property of a date/date-time
**Describe the bug**
I can not access the `weekday` property of a date/date-time value. 

**To Reproduce**
Expression `date("2018-12-10").weekday` fails to evaluate but should return `1`.

**Expected behavior**
A date/date-time value has a property `weekday`. The property is a number value `[1..7]` where `1` is Monday and `7` is Sunday.

**Environment**
* FEEL engine version: 1.11.2

kaaquist:
Signed-off-by: Kasper Aaquist Johansen <kasperaaquist@gmail.com>

## Description

* added the weekday Builtin function to date and datetime. 

closes #160

### DIFF
--- a/docs/develop/feel-expression.md
+++ b/docs/develop/feel-expression.md
@@ -364,15 +364,23 @@ Values of type date, time, date-time and duration have special properties to acc
 ```js
 date("2017-03-10").year                   
 date("2017-03-10").month                
-date("2017-03-10").day                    
+date("2017-03-10").day
+date("2017-03-10").weekday
 
 time("11:45:30+02:00").hour            
 time("11:45:30+02:00").minute         
 time("11:45:30+02:00").second        
 time("11:45:30+02:00").time offset   
 
-date and time("2017-03-10T11:45:30+02:00")    
-// all properties of date and time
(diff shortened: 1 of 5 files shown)


saig0 (changes requested):
@kaaquist thank you for your contribution :cookie: 

In the specification, `weekday` is defined as a property of a date/date-time value. This is different to a built-in function. From the spec:

image

Currently, these properties are defined here: https://github.com/camunda/feel-scala/blob/master/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala#L904-L992

kaaquist:
Sorry about that. I will correct that.

kaaquist:
I have corrected the mistake, and added the new implementation. 👍

saig0 (changes requested):
Looks good. Thank you :cake: 

Please add the new property to the documentation here: https://github.com/camunda/feel-scala/blob/master/docs/develop/feel-expression.md#special-properties

saig0 (approved):
Super :tada:

PR #198, opened 2020-10-03
Add instance of any feature

### LINKED ISSUE #157: instance of Any doesn't work
**Describe the bug**
I can not check if a value is of type `Any` using `instance of`.

**To Reproduce**
Expression `x instance of Any` fails to evaluate. 

**Expected behavior**
`Any` is the upper type of all other types. The expression `x instance of Any` should always return `true` if `x` is not `null`. 

**Environment**
* FEEL engine version: 1.11.2

kaaquist:
## Description

* Add check for instance of Any 

closes #157

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -122,10 +122,8 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        println(
-          s" !!! x: ${x.getClass}, value of x: $x, typeName: ${typeName.getClass}, value of typeName: $typeName :: ${eval(x)}")
         typeName match {
-          case "Any" => ValBoolean(true)
+          case "Any" if eval(x) != ValNull => ValBoolean(true)
           case _ =>
             withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
         }
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -122,11 +122,12 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        typeName match {
-          case "Any" if eval(x) != ValNull => ValBoolean(true)
-          case _ =>
-            withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
-        }
+        withValOrNull(eval(x), x => {
+          typeName match {
+            case "Any" if x != ValNull => ValBoolean(true)
+            case _ => withType(x, t => ValBoolean(t == typeName))
+          }
+        })
 
       // context
       case Ref(names)               => ref(context.variable(names.head), names.tail)
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -122,10 +122,10 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        withValOrNull(eval(x), x => {
+        withVal(eval(x), x => {
           typeName match {
             case "Any" if x != ValNull => ValBoolean(true)
-            case _ => withType(x, t => ValBoolean(t == typeName))
+            case _                     => withType(x, t => ValBoolean(t == typeName))
           }
         })
 
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -122,7 +122,13 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        println(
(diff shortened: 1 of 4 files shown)


src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -122,7 +122,13 @@ class FeelInterpreter {
...
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        println(
+          s" !!! x: ${x.getClass}, value of x: $x, typeName: ${typeName.getClass}, value of typeName: $typeName :: ${eval(x)}")
+        typeName match {
+          case "Any" => ValBoolean(true)
+          case _ =>
+            withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        }
```
saig0, line 131: :+1:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -122,7 +122,13 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        println(
+          s" !!! x: ${x.getClass}, value of x: $x, typeName: ${typeName.getClass}, value of typeName: $typeName :: ${eval(x)}")
```
saig0, line 126: Please remove the log statement :smile:

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -227,10 +227,4 @@ class BuiltinNumberFunctionsTest
     eval(" odd(-2)") should be(ValBoolean(false))
   }
 
-  "A even() function" should "return true if number is even" in {
-
-    eval(" even(5) ") should be(ValBoolean(false))
-    eval(" even(2) ") should be(ValBoolean(true))
-  }
-
```
saig0, line 235: I guess this change came from a different PR.

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
```diff
@@ -60,6 +60,12 @@ class InterpreterExpressionTest
     eval("x instance of string", Map("x" -> 0)) should be(ValBoolean(false))
   }
 
+  it should "be a instance of Any should always pass" in {
+    eval("x instance of Any", Map("x" -> "yes")) should be(ValBoolean(true))
+    eval("x instance of Any", Map("x" -> 1)) should be(ValBoolean(true))
+    eval("x instance of Any", Map("x" -> true)) should be(ValBoolean(true))
+  }
```
saig0, line 67: :+1: 

Please add one more case for `null instance of Any`.

saig0 (changes requested):
Thank you for your contribution @kaaquist :fish_cake: 

Looks good. Just a bit clean up before I can merge it :+1:

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
kaaquist: Added extra test and fixed check for `null`.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
kaaquist: When I merged my fork with the "new changes to upstream/master, this here test got added two times in my code base.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -122,7 +122,11 @@ class FeelInterpreter {
       case In(x, test) =>
         withVal(eval(x), x => eval(test)(context + (inputKey -> x)))
       case InstanceOf(x, typeName) =>
-        withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        typeName match {
+          case "Any" if eval(x) != ValNull => ValBoolean(true)
+          case _ =>
+            withVal(eval(x), x => withType(x, t => ValBoolean(t == typeName)))
+        }
```
saig0, line 129: We can optimize the expression a bit to evaluate `x` only once. 

```suggestion
        withValOrNull(eval(x), x => {
          typeName match {
            case "Any" if x != ValNull => ValBoolean(true)
            case _ => withType(x, t => ValBoolean(t == typeName))
          }
        })
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
saig0: Hmm. :thinking: I'm afraid that the test will be deleted when the PR is merged.

saig0 (changes requested):
Thanks :+1: 

Just one minor improvement.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
kaaquist: Beautiful 😄

kaaquist:
@saig0  after approving the change.. I now get an error: 
```
[ERROR] /Users/kasper/developer/github.com/kaaquist/feel-scala/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala:125: too many arguments (2) for method withValOrNull: (x: org.camunda.feel.syntaxtree.Val)org.camunda.feel.syntaxtree.Val
```

Do I need to rebase from mater to get the `withValOrNull` function you proposed to use?

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
kaaquist: I have added this here test again.. I don't know why it got lost in the first place. But now it is back.

kaaquist:
Ok, I used `withVal` instead of the proposed `withValOrNull` function. Now it compiles and passed all tests 👍

saig0 (approved):
Thanks :tada:

PR #199, opened 2020-10-09
fix: fixed abs() function to be invoked with named parameter "n"

### LINKED ISSUE #164: abs() function can not be invoked with named parameter "n"
**Describe the bug**
The function `abs()` can not be invoked with a named parameter `n`.

With DMN 1.3, the parameter name of the `abs()` function changed from `number` to `n`.

**To Reproduce**
Expression `abs(n: 1)` fails to evaluate but should return `1`

**Expected behavior**
The `abs()` function can be invoked either with a named parameter `n` or `number`.

**Environment**
* FEEL engine version: 1.11.2

daniel-shuy:
## Description

* The `abs()` can now be invoked either with a named parameter `n` or `number`

## Related issues

closes #164

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
@@ -12,7 +12,7 @@ object NumericBuiltinFunctions {
     "decimal" -> List(decimalFunction, decimalFunction3),
     "floor" -> List(floorFunction),
     "ceiling" -> List(ceilingFunction),
-    "abs" -> List(absFunction),
+    "abs" -> List(absFunctionNumber, absFunctionN),
     "modulo" -> List(moduloFunction),
     "sqrt" -> List(sqrtFunction),
     "log" -> List(logFunction),
@@ -71,10 +71,14 @@ object NumericBuiltinFunctions {
     }
   }
 
-  private def absFunction =
-    builtinFunction(params = List("number"), invoke = {
+  private def absFunction(paramName: String) =
+    builtinFunction(params = List(paramName), invoke = {
       case List(ValNumber(n)) => ValNumber(n.abs)
     })
+
+  private def absFunctionNumber = absFunction("number")
+
+  private def absFunctionN = absFunction("n")
 
   private def moduloFunction =
     builtinFunction(params = List("dividend", "divisor"), invoke = {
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
```diff
@@ -12,7 +12,7 @@ object NumericBuiltinFunctions {
     "decimal" -> List(decimalFunction, decimalFunction3),
     "floor" -> List(floorFunction),
     "ceiling" -> List(ceilingFunction),
-    "abs" -> List(absFunction),
+    "abs" -> List(absFunctionNumber, absFunctionN),
```
saig0, line 15: We can declare the functions directly here.

```suggestion
    "abs" -> List(absFunction(paramName = "number"), absFunction(paramName = "n")),
```

src/main/scala/org/camunda/feel/impl/builtin/NumericBuiltinFunctions.scala
```diff
@@ -71,11 +71,15 @@ object NumericBuiltinFunctions {
...
-    builtinFunction(params = List("number"), invoke = {
+  private def absFunction(paramName: String) =
+    builtinFunction(params = List(paramName), invoke = {
       case List(ValNumber(n)) => ValNumber(n.abs)
     })
 
+  private def absFunctionNumber = absFunction("number")
+
+  private def absFunctionN = absFunction("n")
+
```
saig0, line 82: If the functions are declared in the list then we can remove them here.

```suggestion
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -180,6 +180,18 @@ class BuiltinNumberFunctionsTest
     eval(" abs(-10) ") should be(ValNumber(10))
   }
 
+  "abs() function" should "be invoked with named parameter number" in {
```
saig0, line 183: We can use the `it` statement to group the test cases.

```suggestion
  it should "be invoked with named parameter number" in {
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinNumberFunctionTest.scala
```diff
@@ -180,6 +180,18 @@ class BuiltinNumberFunctionsTest
...
+
+    eval(" abs(number: 1) ") should be(ValNumber(1))
+    eval(" abs(number: -1) ") should be(ValNumber(1))
+  }
+
+  "abs() function" should "be invoked with named parameter n" in {
```
saig0, line 189: We can use the `it` statement to group the test cases.

```suggestion
  it should "be invoked with named parameter n" in {
```

saig0 (changes requested):
Thank you for your contribution @daniel-shuy :cookie: 

Looks good. I like your solution :+1: 
I have just some suggestions for minor improvements. Please have a look :)

daniel-shuy:
Done! Should I squash my commits?

saig0 (approved):
Thanks @daniel-shuy :tada:

PR #202, opened 2020-10-16
Publish new artifact which shades the Scala Library 

### LINKED ISSUE #149: Publish additional artifact with shaded Scala Library
**Is your feature request related to a problem?**

* The FEEL Engine is compatible with a specific version of the Scala Library
* When my project already uses an incompatible version of the Scala Library, I cannot use the FEEL Engine in my project

An additional artifact is published which shades the Scala Library.

**Related issues**

Camunda BPM: https://jira.camunda.com/browse/CAM-12136

tasso94:
## Description

The Scala Library is shaded inside a new artifact with the classifier `scala-shaded`

## Related issues

closes #149

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -251,7 +251,7 @@
               <shadedClassifierName>scala-shaded</shadedClassifierName>
               <createDependencyReducedPom>true</createDependencyReducedPom>
               <promoteTransitiveDependencies>false</promoteTransitiveDependencies>
-              <createSourcesJar>false</createSourcesJar>
+              <createSourcesJar>true</createSourcesJar>
               <keepDependenciesWithProvidedScope>true</keepDependenciesWithProvidedScope>
               <artifactSet>
                 <excludes>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -235,6 +236,40 @@
...
+            <phase>package</phase>
+            <goals>
+              <goal>shade</goal>
+            </goals>
+            <configuration>
+              <shadedArtifactAttached>true</shadedArtifactAttached>
+              <shadedClassifierName>scala-shaded</shadedClassifierName>
+              <createDependencyReducedPom>true</createDependencyReducedPom>
+              <promoteTransitiveDependencies>false</promoteTransitiveDependencies>
+              <createSourcesJar>false</createSourcesJar>
```
tasso94, line 254:
```suggestion
              <createSourcesJar>true</createSourcesJar>
```
We should probably create a sources jar of the scala-shaded artifact.

saig0 (approved):
Looks good as far as I can tell :)

pom.xml
koevskinikola: :+1: I think it would be good to have them as well.

koevskinikola (approved):
Looks good! You can flip the sources flag if you like. I think it would be useful to have them.

PR #213, opened 2020-11-18
fix(ObjectContext): respect boolean accessors prepended with "is"

### LINKED ISSUE #212: A boolean accessor prepended with "is" is not resolved
### Describe the bug

When evaluating an expression with an `ObjectContext`, a boolean accessor prepended with "is" is not included in the FEEL context. Only accessors prepended with "get" are taken into account. 

### Steps to reproduce

Evaluate the following FEEL expression `bar.disabled = false` and pass an object variable based on the following type:
```scala
  class MyScalaType {

    private val disabled = false

    def isDisabled = disabled

  }
```

### Expected behavior

The expression returns `true`. 

### Observed behavior

The attribute `disabled` does not exist in the context.

### Environment
* FEEL engine version: 1.12.3
* Affects: 
  * Camunda BPM: 7.13.7, 7.14.1: https://jira.camunda.com/browse/CAM-12577

### Context

The customer uses Lombok

tasso94:
### Description

In `ObjectContext`, respect boolean accessors prepended with "is".

### Related issues

related to CAM-12577, closes #212

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/ObjectContext.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/ObjectContext.scala
@@ -29,14 +29,21 @@ case class ObjectContext(obj: Any) extends Context {
   override val variableProvider = new VariableProvider {
     override def getVariable(name: String): Option[Any] = {
 
-      val field = obj.getClass.getFields find (f => f.getName == name)
+      val objClass = obj.getClass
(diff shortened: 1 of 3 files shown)


tmetzke (approved):
Looks good to me, nice extensive test coverage! 👍 
I also like that we use built-in functions like `capitalize` instead of doing that manually 😃

PR #222, opened 2020-12-21
Fastparse

nightscape:
## Description

* Switched from scala-parser-combinators to fastparse
* Commented out failing tests with rather surprising behavior.
* Temporarily removed all cuts (points beyond which the parser cannot backtrack) to get tests to pass

## Related issues

Prerequisite for https://github.com/camunda/feel-scala/pull/221

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -16,453 +16,551 @@
  */
 package org.camunda.feel.impl.parser
 
+import fastparse.JavaWhitespace._
+import fastparse._
 import org.camunda.feel._
 import org.camunda.feel.syntaxtree._
 
 import scala.util.Try
-import scala.util.parsing.combinator.JavaTokenParsers
-
 object FeelParser extends JavaTokenParsers {
 
-  def parseExpression(exp: String): ParseResult[Exp] = parseExp(expression, exp)
+  def parseExpression(exp: String): Parsed[Exp] =
+    parse(exp, fullExpression(_))
+  private def fullExpression[_: P] = "" ~ expression ~ End
+  def parseUnaryTests(expression: String): Parsed[Exp] =
+    parse(expression, fullUnaryExpression(_))
+  private def fullUnaryExpression[_: P] = P("" ~ unaryTests ~ End)
 
-  def parseUnaryTests(expression: String): ParseResult[Exp] =
-    parseExp(unaryTests, expression)
-
-  private def parseExp[T](parser: Parser[T], exp: String) =
-    parseAll(parser, exp)
-
-  // override to ignore comment '// ...' and '/* ... */'
-  protected override val whiteSpace =
-    """(\s|//.*|(?m)/\*(\*(?!/)|[^*])*\*/)+""".r
-
-  private lazy val reservedWord: Parser[String] = ("null\\b".r
-    | "true\\b".r | "false\\b".r
-    | "function\\b".r
-    | "if\\b".r | "then\\b".r | "else\\b".r
-    | "for\\b".r
-    | "between\\b".r
-    | "instance\\b".r | "of\\b".r)
+  private def reservedWord[_: P]: P[String] =
+    P[String](
+      StringIn("null",
+               "true",
+               "false",
+               "function",
+               "if",
+               "then",
+               "else",
+               "for",
+               "between",
+               "instance",
+               "of").!)
 
   // list of built-in function names with whitespaces
   // -- other names match the 'function name' pattern
-  private lazy val builtinFunctionName: Parser[String] = ("date and time"
-    | "years and months duration"
-    | "string length"
-    | "upper case"
-    | "lower case"
-    | "substring before"
-    | "substring after"
-    | "starts with"
-    | "ends with"
-    | "list contains"
-    | "insert before"
-    | "index of"
-    | "distinct values"
-    | "get entries"
-    | "get value"
-    | "is defined"
-    | "day of year"
-    | "day of week"
-    | "month of year"
-    | "week of year")
+  private def builtinFunctionName[_: P]: P[String] = P[String](
+    StringIn(
+      "date and time",
+      "years and months duration",
+      "string length",
+      "upper case",
+      "lower case",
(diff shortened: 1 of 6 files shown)


nightscape:
@saig0 it would be great if you could have a first look.
I'd be especially interested, if [the tests that are currently commented out](https://github.com/camunda/feel-scala/pull/222/files#diff-47f12e411a58c6f00e2e0a8709371702726ac241b731d910abf44d98ce84c6bcR151) (unary tests that compare a provided Boolean value to a literal `true` or `false` without a comparison operator) still make sense.

saig0:
@nightscape awesome, thank you for your contribution :tada: 

Please note that I'll need some time to review the changes. I didn't use FastParse before and I don't have much time these days :christmas_tree:  

Regarding:

> I'd be especially interested, if the tests that are currently commented out (unary tests that compare a provided Boolean value to a literal true or false without a comparison operator) still make sense.

Yes, the behavior makes sense according to the specification :sweat_smile: 
A unary tests is used in an input entry of a decision table. Usually, it doesn't contain a comparison operator (to make the condition more readable in the decision table). If it is just a literal then it implicitly compares the values (i.e. `true` => `input = true`)
In this case, it looks a bit weird for two boolean values but it is correct.  

Regarding priorities and scoping, is this PR required for the cross-compiling to JS? Or, it is possible and reasonable to migrate to FastParse later?

nightscape:
I think it is required or at least very helpful for the JS migration.
You mentioned that you tried a Scala-JS port a while ago and you ran into problems with Regexes. The same happened for me and I know that Regexes are one area where Scala-JS behavior differs from JVM Scala.
Maybe those problems could be fixed, but as you mentioned you wanted to migrate to fastparse (I also think that's a good idea), I reasoned it would be a good idea to sequence it this way around.
Using fastparse actually solved these errors and the remaining problems with Scala-JS have to do with reflection.

fastparse is actually very similar to scala-parser-combinators, I did most of the migration using search and replace.
The main differences are whitespace handling (which is mostly automatic in fastparse), capturing (which has to be done explicitly using `.!` in fastparse) and pointcuts (which prevent the parser from backtracking).

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
```diff
@@ -147,6 +147,8 @@ class InterpreterUnaryTest
     evalUnaryTests("c", """ "a","b" """) should be(ValBoolean(false))
   }
 
+  /*
+  TODO: These tests currently fail, but the specified behavior seems rather surprising anyway
```
saig0, line 151: These tests are correct and must be stay enabled. 

Example use case:
* a decision table with an input that is named "is valid"
* the input has an input expression "contract.isValid"
* a rule with an input entry with the unary-test `true`
* the rule should only match if `contract.isValid` is equal to `true`

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -16,475 +16,505 @@
...
-  )
+  private def positiveUnaryTest[_ : P]: P[Exp] =
+    P[Exp](
+      P("null") ^^^ InputEqualTo(ConstNull) |
+        // TODO: This makes it seemingly impossible for fastparse to backtrack its way out
+        //  when it should be using the expression.map part.
+        //  From the comment above it should not be necessary.
+        // simplePositiveUnaryTest |
+        expression.map(UnaryTestExpression)
+    )
```
saig0, line 162: It seems that the backtracking fails when evaluating the `identifier` parser :thinking: 

However, we need to parse `simplePositiveUnaryTest` before `expression` to fulfil the boolean test cases in `InterpreterUnaryTest`. Otherwise, a boolean value is not parsed as an `InputEqualTo(ValBoolean())` expression. Instead, it is parsed as `UnaryTestExpression(ValBoolean())` and evaluates to `true` if the boolean value is `true`. See my comment on `InterpreterUnaryTest`.

We can make it work by modifying the `simpleValue` parser:
```
private def simpleValue[_: P] =
    P(
      simpleLiteral |
        !(functionInvocation | builtinFunctionInvocation) ~ qualifiedName.map(
          Ref(_)) |
        qmark
    )
```

saig0 (changes requested):
Nice work @nightscape :+1: Thanks a lot :cake: 

I added a few minor comments and a suggestion to fix the failing test cases. 
I see more ways to improve the `FeelParser` but this can be done later.

nightscape (commented):
@saig0 I think I addressed all your points and also

- reintroduced most cuts except for the one in `contextEntry` which breaks tests
- removed comments
- rebased everything to a single commit.

From my side this is good to merge :partying_face:

nightscape:
Hmm, I don't know why it's formatting code differently on my machine.
I could minimize the diff a little more if you can tell me how you do the automatic formatting.

nightscape:
@saig0 would it be an option to update the `mvn_scalafmt` plugin, add a `.scalafmt.conf` file with your preferred formatting and then run `scalafmt` once on the entire project?
That should (more or less) guarantee that contributors generate the same formatting and you don't get accidental changes.

If you need a `.scalafmt.conf`, here is my preferred one which minimizes the amount of accidentally changed lines in Git due to alignment:
```
version = 2.7.5
style = default
maxColumn = 120
continuationIndent.defnSite = 2
continuationIndent.callSite = 2
align.preset = none
danglingParentheses.preset = true
optIn.configStyleArguments = false
docstrings = ScalaDoc
spaces.beforeContextBoundColon = Always
rewrite.rules = [SortImports]
```

saig0:
The code should be formatted when running `mvn clean install`.

I would like update the `mvn-scalafmt` plugin. But it requires Maven 3.6 which is currently not available in our CI. See
* #127
* #136

nightscape:
@saig0 I ran `mvn compile` which afaict also ran the code formatting, but the formatting still turned out slightly different.
Would adding a `.scalafmt.conf` be an option? That might reduce some ambiguities in how scalafmt formats the code.
If not, I could also manually undo the formatting changes.

saig0 (approved):
Looks good :tada: 

Let's introduce the `.scalafmt.conf` in a separate PR :)

PR #237, opened 2021-02-19
Simplify the parser

saig0:
## Description

* use common pattern from the parser documentation, for example:
  * replace the custom `^^^` operator with `map()` 
  * replace the custom `chainl1` method with `flapMap()` 
  * use `AnyChar` with negation 
* remove useless logging for invalid date-time formats

## Related issues

Follow-up of #222

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -22,17 +22,29 @@ import org.camunda.feel._
 import org.camunda.feel.syntaxtree._
 
 import scala.util.Try
-object FeelParser extends JavaTokenParsers {
 
-  def parseExpression(exp: String): Parsed[Exp] =
-    parse(exp, fullExpression(_))
-  private def fullExpression[_: P] = "" ~ expression ~ End
+/**
+  * The parser is written following the FEEL grammar definition in the DMN specification.
+  *
+  * In order to understand how the parser works, it is recommended to read the documentation first:
+  * [[https://www.lihaoyi.com/fastparse]]. Additional resources:
+  * [[https://www.lihaoyi.com/post/EasyParsingwithParserCombinators.html]],
+  * [[https://www.lihaoyi.com/post/BuildyourownProgrammingLanguagewithScala.html]]
+  */
+object FeelParser {
+
+  def parseExpression(expression: String): Parsed[Exp] =
+    parse(expression, fullExpression(_))
+
   def parseUnaryTests(expression: String): Parsed[Exp] =
     parse(expression, fullUnaryExpression(_))
-  private def fullUnaryExpression[_: P] = P("" ~ unaryTests ~ End)
 
-  private def reservedWord[_: P]: P[String] =
-    P[String](
+  private def fullExpression[_: P] = P(Start ~ expression ~ End)
+
+  private def fullUnaryExpression[_: P] = P(Start ~ unaryTests ~ End)
+
+  private def reservedWord[_: P] =
+    P(
       StringIn("null",
                "true",
                "false",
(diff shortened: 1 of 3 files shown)


src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -18,21 +18,34 @@ package org.camunda.feel.impl.parser
 
 import fastparse.JavaWhitespace._
 import fastparse._
+import fastparse.internal.Logger
```
korthout, line 21: Considering that all uses of the logger have been removed, this may also be removed again

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -18,21 +18,34 @@ package org.camunda.feel.impl.parser
...
-  def parseExpression(exp: String): Parsed[Exp] =
-    parse(exp, fullExpression(_))
-  private def fullExpression[_: P] = "" ~ expression ~ End
+/**
+  * In order to understand how the parser works, it is recommended to read the documentation first:
+  * [[https://www.lihaoyi.com/fastparse]]. Additional resources:
+  * [[https://www.lihaoyi.com/post/EasyParsingwithParserCombinators.html]],
+  * [[https://www.lihaoyi.com/post/BuildyourownProgrammingLanguagewithScala.html]]
+  */
+object FeelParser {
```
korthout, line 35: Maybe also add a link to the grammar file ;)

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -43,92 +56,133 @@ object FeelParser extends JavaTokenParsers {
...
   // -- other names match the 'parameter name' pattern
-  private def builtinFunctionParameterNames[_: P]: P[String] =
-    P[String](
-      StringIn("start position", "grouping separator", "decimal separator").!)
+  private def builtinFunctionParameterNames[_: P] =
+    P(StringIn("start position", "grouping separator", "decimal separator"))
 
-  private def identifier[_: P]: P[String] = P(!(reservedWord ~~ !ident) ~ ident)
+  // an identifier which is not a reserved word. however, it can contain a reserved word.
+  private def identifier[_: P] = P(reservedWord.? ~~ javaLikeIdentifier)
```
korthout, line 95: The reservedWord is no longer repeated. What if I write `foobarX` (where both foo and bar are reserved), is this then failing? Also, how does this allow for something like `prefixfooX` (where foo is reserved)?

korthout (commented):
I mostly have questions due to my lack of knowledge, but please have a look.

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -43,92 +56,133 @@ object FeelParser extends JavaTokenParsers {
...
+        Character.isJavaIdentifierPart,
+        0)
+    )
 
-  // Java-like string literal: ("\""+"""([^"\x00-\x1F\x7F\\]|\\[\\'"bfnrt]|\\u[a-fA-F0-9]{4})*"""+"\"")
-  // modification: allow '\'
   private def stringLiteralWithQuotes[_: P]: P[String] =
-    //P[String](""""([^"\x00-\x1F\x7F]|\\u[a-fA-F0-9]{4})*"""".r.map(_.replaceAll("\"", "")))
-    P[String]("\"" ~~ CharsWhile(_ != '"', 0).! ~~ "\"")
+    P("\"" ~~ (!"\"" ~~ AnyChar.!).repX ~~ "\"")
```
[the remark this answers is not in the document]
saig0: `AnyChar` also captures quotes. Without the negative lookahead, it would consume the quote. I took this example from the docs.

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -43,92 +56,133 @@ object FeelParser extends JavaTokenParsers {
...
+
   // 2 c)
-  private def expression3[_: P] = P(conjunction)
+  private def expression3[_: P]: P[Exp] = P(conjunction)
+
   // 2 d)
   private def expression4[_: P]: P[Exp] =
-    P[Exp](expression5.flatMap(optionalComparison))
+    P(expression5.flatMap(x =>
+      comparison(x).?.map(_.fold(x)(comparisonExpr => comparisonExpr))))
```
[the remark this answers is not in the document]
saig0: Yes, this part is a bit tricky :smile:

`flatMap()` is an optimization. It could be also written as `expression5 ~ comparison | expression5`. With the `flatMap()`, we need to parse `expression5` only once.

`_.fold(x)(...)` is applied on the `Option` of `comparison(x).?`. It returns `x` if no comparison is present. Otherwise, it returns the comparison. See the Scala docs [here](https://www.scala-lang.org/api/current/scala/Option.html#fold[B](ifEmpty:=%3EB)(f:A=%3EB):B).

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
saig0: The parser accepts two reserved words or a reserved word with a prefix as a valid name. Only reserved words are not allowed. 

If it parses `reservedWord` then it must be followed by `javaLikeIdentifier` (any additional char). 
If it has a prefix then it is not parsed as `reservedWord`.

saig0:
@korthout thank you for your review :+1: 

Does this answer your questions?

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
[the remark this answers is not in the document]
korthout: Thanks, the docs really cleared it up for me.

korthout (approved):
@saig0 Thanks for the answers. It really helps.

LGTM 💯

PR #239, opened 2021-02-19
New context built-in functions put()/put all()

### LINKED ISSUE #169: New function to add new entries to a context
Currently, there is no function to append entries to an existing context. A context can only be created statically. 

In an expression, sometimes you need to append one or more entries to a context value variable. Another use case is to merge two context values together into a new context which has all entries of both contexts.

A new built-in function `put()` as equivalent to `append()` but for a context.

```
put({}, "x", 1) = {x: 1}

put({x:1}, {y:2}) = {x: 1, y: 2}
```

**Related issues**

Currently, Zeebe contains a similar function that is used for the (output) variable mappings. This can be replaced when the function exists in the FEEL engine itself. (https://github.com/zeebe-io/zeebe/blob/develop/expression-language/src/main/scala/io/zeebe/el/impl/feel/FeelFunctionProvider.scala#L25-L43)

saig0:
## Description

* add new context built-in functions `put()` and `put all()`

In Zeebe, we can replace the custom `appendTo(x, y)` function by `if (x = null) then y else put all(x, y)`.

## Related issues

closes #169

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -68,7 +68,8 @@ object FeelParser extends JavaTokenParsers {
       "day of year",
       "day of week",
       "month of year",
-      "week of year"
+      "week of year",
+      "put all"
     ).!
   )
 
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -16,6 +16,7 @@
  */
 package org.camunda.feel.impl.builtin
 
+import org.camunda.feel.context.Context.StaticContext
 import org.camunda.feel.impl.FeelIntegrationTest
 import org.scalatest.FlatSpec
 import org.scalatest.Matchers
@@ -47,7 +48,7 @@ class BuiltinContextFunctionsTest
       Map("key" -> ValString("foo"), "value" -> ValNumber(123)))
   }
 
-  it should "return empty list if emtpy" in {
+  it should "return empty list if empty" in {
 
     eval(""" get entries({}) """) should be(ValList(List()))
   }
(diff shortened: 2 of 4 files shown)


src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -68,7 +68,8 @@ object FeelParser extends JavaTokenParsers {
       "day of year",
       "day of week",
       "month of year",
-      "week of year"
+      "week of year",
+      "put all"
```
korthout, line 72: Although the comment on this says it only requires function names with whitespaces. I wonder whether you also need to add `"put"` somewhere in the parsers. For example, `date` is also explicitly defined in this class, `time` and `not` as well, but `decimal` is not.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -62,4 +63,116 @@ class BuiltinContextFunctionsTest
...
+    eval(""" put all({x:1}, {y:2}, {z:3}) """) should be(
+      ValContext(
+        StaticContext(variables =
+          Map("x" -> ValNumber(1), "y" -> ValNumber(2), "z" -> ValNumber(3)))
+      ))
+  }
+
+  it should "add a nested context" in {
+
+    eval(""" put all({x:1}, {y:{z:2}}) = {x:1, y:{z:2} } """) should be(
```
korthout, line 169: Why the predicate here instead of checking the entries like in all other tests?

korthout (approved):
Great work! Easy to understand PR and easy to read tests.

One thing to note I miss some tests concerning the overwriting order for values of the same key in the contexts passed into the `put all` function, but since it is also not documented this should not hold back the functionality. However, it might be worth to document, and in that case should also be tested.

I only have some really minor stuff, so: LGTM 🦕

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
saig0: The function name `put` can be parsed by the `name` parser. We don't need to handle it explicitly. It is different for names with whitespace because they aren't allowed as function names.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
saig0: The nested structure is harder to test. The inner `context` is an interface with different implementations. 
We could extract the variables out of it in the test case. But it is easier to read if we check it in the expression itself.

PR #263, opened 2021-04-27
Parse nested path and filter expressions

### LINKED ISSUE #258: complex expressions on lists not parsed correctly since 1.13.0
**Describe the bug**
Since 1.13.0, complex expressions with filters on lists are not correctly parsed by FeelEngine.

**To Reproduce**
1. create following objects
``` kotlin
data class CaseStationFile(
    val id: String,
    val applicantsPersonalData: List<ApplicantsPersonalData> = emptyList(),
)

data class ApplicantsPersonalData(
    val role: String,
    val documents: List<Document> = emptyList(),
)

data class Document(
    val type: String,
    val isValidated: Boolean,
)
```
2. create an instance of CaseStationFile with values below and add it in a Map with key `caseFile`
```kotlin
val caseFile = CaseStationFile("id", listOf(ApplicantsPersonalData("primaryBorrower", listOf(Document("idCard", true)))))
val variables = mapOf("caseFile" to caseFile)
```
3. pass the following expression and the Map to FeelEngine.evalExpression:
``` kotlin
val result = feelEngine.evalExpression("((caseFile.applicantsPersonalData[role=\"primaryBorrower\"])[1].documents[type=\"idCard\"])[1].isValidated", variables)
```

**Expected behavior**
result of evalExpression should be equal to true

It was OK with 1.12.4 and I get following parsing error since 1.13.0:
```
failed to parse expression '((caseFile.applicantsPersonalData[role="primaryBorrower"])[1].documents[type="idCard"])[1].isValidated': Expected (StringIn("<=", ">=", "<", ">", "!=", "=") | "between" | "in" | "and" | "or" | ".."):1:87, found ")[1].isVal"
```

**Environment**
* FEEL engine version: [1.13.1]

saig0:
## Description

* after replacing the parser library in `1.13.0`, some expressions with nested filter and path access can not be parsed anymore (e.g. `([1,2,3,4][item > 2])[1]`, `([{x:{y:1}},{x:{y:2}},{x:{y:3}}]).x[2].y`)
* introduce new parser combinators to parser nested filter and path expressions that follow an expression in brackets
  * instead of using a recursive parsing style as it is currently the case, use a continuation parsing style
  * this style fits better for parsing an expression from left to right and doesn't require limitations to avoid endless recursions
  * since this style seems promising to reduce some limitations of the parser, I plan to roll it out more step-by-step 

## Related issues

closes #258

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -17,6 +17,7 @@
 package org.camunda.feel.impl.parser
 
 import fastparse.JavaWhitespace._
+import fastparse.Parsed.Success
 import fastparse._
 import org.camunda.feel._
 import org.camunda.feel.syntaxtree._
@@ -118,7 +119,29 @@ object FeelParser {
 
   // 2 a)
   private def textualExpression[_: P]: P[Exp] =
-    P(functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+    P(expressionInParentheses | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+
+  private def expressionInParentheses[_: P]: P[Exp] =
+    P("(" ~ expression ~ ")").flatMap(continuation)
+
+  // optimize parsing the expression from left to right by using `flapMap()`
+  // `flatMap()` takes the parsed (left) part of the expression and continues with the (right) part
+  // it does backtracking if `flatMap()` is not successful
+  private def continuation[_: P](base: Exp): P[Exp] = P(
+    continuationExpression(base).flatMap(continuation) |
+      End.map(_ => base)
+  )
+
+  private def continuationExpression[_: P](base: Exp): P[Exp] =
+    filterContinuation(base) | pathContinuation(base)
+
+  private def filterContinuation[_: P](base: Exp): P[Exp] =
+    P("[" ~ expression ~ "]")
+      .map(filter => Filter(base, filter))
+
+  private def pathContinuation[_: P](base: Exp): P[Exp] =
+    P(("." ~ name).rep(1))
+      .map(ops => ops.foldLeft(base)(PathExpression))
 
   // 2b)
   private def expression2[_: P]: P[Exp] = P(disjunction)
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -118,7 +119,29 @@ object FeelParser {
 
   // 2 a)
   private def textualExpression[_: P]: P[Exp] =
-    P(functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+    P(expressionInBrackets | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+
+  private def expressionInBrackets[_: P]: P[Exp] =
```
korthout, line 124: I think the word brackets is generally used specifically for square brackets `[ ]`. Instead, the word parentheses is used for round brackets `( )`. See [bracket](https://en.wikipedia.org/wiki/Bracket).

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -118,7 +119,29 @@ object FeelParser {
...
+    continuationExpression(base).flatMap(continuation) |
+      End.map(_ => base)
+  )
+
+  private def continuationExpression[_: P](base: Exp): P[Exp] =
+    filterContinuation(base) | pathContinuation(base)
+
+  private def filterContinuation[_: P](base: Exp): P[Exp] =
+    P("[" ~ expression ~ "]")
+      .map(filter => Filter(base, filter))
```
korthout, line 140: This seems kind of duplicate of `filteredExpression9`. Does this mean the other filterExpressions can be removed or replaced in the future?

korthout (approved):
I really like the new style. It reads much easier comparing `filterContinuation` to something like `filteredExpression9`.

LGTM already, just a question and a naming suggestion.

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
saig0: Yes. I plan to apply the new style to more places. If this is done then I'll remove the duplications :+1: 
Currently, I want to keep the changes small.

PR #269, opened 2021-05-06
WIP: New Feature: add a function that extracts text based on regex

younes-io:
Adding a function that extracts all occurrences of a regex in an input string

## Description

* Added the function `extractFunction` that takes two arguments: an `input` string and a `pattern`, then returns all occurrences of the pattern
* Added the function `extractFunction3` that takes three arguments: an `input` string, a `pattern`, and `flags` then returns all occurrences of the pattern

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
@@ -19,7 +19,8 @@ object StringBuiltinFunctions {
     "starts with" -> List(startsWithFunction),
     "ends with" -> List(endsWithFunction),
     "matches" -> List(matchesFunction, matchesFunction3),
-    "split" -> List(splitFunction)
+    "split" -> List(splitFunction),
+    "extract" -> List(extractFunction, extractFunction3)
   )
 
   private def substringFunction = builtinFunction(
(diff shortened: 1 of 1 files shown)


saig0:
Hi @younes-io,

thank you for contributing :+1: 

Before starting with a PR, please create an issue first. The issue is used as documentation and to discuss the bug/feature. It should save time for the contributor and the maintainers. 

In this case, it would be good to know what are the benefits over the [replace()](https://camunda.github.io/feel-scala/docs/reference/builtin-functions/feel-built-in-functions-string#replace) function. And examples may help to get your point.

vincentgiraud:
Hello @saig0,
Following your recommendation, I've created https://github.com/camunda/feel-scala/issues/315 to explain the differences with the replace() function along with some examples. As I understand, the replace() function would require to make a negative lookahead in order to replace all that is not the regex with null. Unfortunately I couldn't get it to work and it'd be a pretty counter intuitive logic to set up. Also, it seems that we need to specify the member group ids where we won't not for sure the occurence count.

vincentgiraud:
Hello @younes-io  and @saig0, as I didn't have the authz to commit on top of this, I just created a new pr:  https://github.com/camunda/feel-scala/pull/317

saig0:
closing this PR is favor of #317

PR #302, opened 2021-07-14
Allow parentheses in context literal

### LINKED ISSUE #278: Failed to parse expression with parentheses
**Describe the bug**
The following expression with parentheses can't be parsed:

```
{x:(xs[1])}
```

The same expression without parentheses can be parsed.

image

**To Reproduce**
1. parse the expression `{x:(xs[1])}`
2. verify that the parsing failed

**Expected behavior**
The expression can be parsed with and without parentheses.

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: `1.0.0` https://github.com/camunda-cloud/zeebe/issues/7132

saig0:
## Description

* fix parsing failure if a context literal contains parentheses
* the [cut](https://com-lihaoyi.github.io/fastparse/#Cuts) in the filter expression prevented the parser from backtracking

## Related issues

closes #278

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -483,7 +483,7 @@ object FeelParser {
   // 52
   private def filteredExpression9[_: P]: P[Exp] =
     P(expression9.flatMap(x =>
-      ("[" ~/ expression ~ "]").?.map(_.fold(x)(filterExp =>
+      ("[" ~ expression ~ "]").?.map(_.fold(x)(filterExp =>
         Filter(x, filterExp)))))
 
   // 40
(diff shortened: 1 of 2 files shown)


korthout (approved):
It was clear from the tests that this resolves the problem, but it took me some time to understand why. What happened was that I ran into DMN spec version differences 😅 

52 seemed to refer to the definition of the type in the right hand side of the `instance of` operator, but the code was using this as a filter expression. I thought, perhaps I'm reading it wrong and what is meant here is that the `instance of` can be part of a filter expression (which would make some sense). But, then it hit me: The numbers were wrong. I was looking at DMN 1.3.

Considering that the 1.2 spec states 52 as: `filter expression = expression , "[" , expression , "]" ;`, it makes sense that we shouldn't stop the backtracking here using a cut.

LGTM 👍

saig0:
@korthout thank you for your review 👍 

You're right that the numbers in the comments doesn't match anymore. I'll remove these comments 👍 

Also, I'll move away from the grammar in the specification. I see a more efficient way to parse the grammar. More in further PRs.

PR #303, opened 2021-07-14
Allow chaining of multiple filter expressions

### LINKED ISSUE #279: Fail to parse expression with more than two filters
**Describe the bug**
The engine fails to parse an expression that contains more than two filters, or a combination of multiple filter and path expressions.

For example:
```
x[1][2][3]

[[[1]]][1][1][1]

{x:[[1,2],[3,4]]}.x[2][1]
```

image

**To Reproduce**
1. Parse the expression `[[[1]]][1][1][1]`
2. Verify that the expression failed to parse

**Expected behavior**
An expression with multiple filter and path expressions can be parsed.

The expression `[[[1]]][1][1][1]` should return `1`.

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

saig0:
## Description

* fix parsing failure if a filter expression is chained more than two times (e.g. `x[1][2][3]`)
* extend the previous parser approach of continuations by all expressions that can have a continuation 

## Related issues

closes #279

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -114,10 +114,16 @@ object FeelParser {
     P(textualExpression.rep(1, sep = ",").map(s => ConstList(s.toList)))
 
   private def textualExpression[_: P]: P[Exp] =
-    P(expressionInParentheses | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+    P(expressionWithContinuation | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
 
-  private def expressionInParentheses[_: P]: P[Exp] =
-    P("(" ~ expression ~ ")").flatMap(continuation)
+  private def expressionWithContinuation[_: P]: P[Exp] =
+    P(
+      "(" ~ expression ~ ")" |
+        variableExpression |
+        list |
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -114,10 +114,16 @@ object FeelParser {
     P(textualExpression.rep(1, sep = ",").map(s => ConstList(s.toList)))
 
   private def textualExpression[_: P]: P[Exp] =
-    P(expressionInParentheses | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
+    P(expressionWithContinuation | functionDefinition | forExpression | ifExpression | quantifiedExpression | expression2)
```
korthout, line 117: 🔧  While you're at it, perhaps also replace expression2 with disjunction directly. I believe that improves the understandability of the textualExpression.

korthout (approved):
LGTM 👍 

Please see my suggestion

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
saig0: True. I'll change it in a further PR. I started to rewrite the structure of the parser and how the parser combinators are chained.

PR #304, opened 2021-07-14
Allow unary-tests expressions with some/every

### LINKED ISSUE #280: Fail to parse some/every expression in unary-tests
**Describe the bug**
The engine fails to parse a unary-tests expression that contains a `some` or `every` expression.

```
every x in ? satisfies x > 5
```

image

**To Reproduce**
1. Parse the expression `every x in ? satisfies x > 5` 
2. Verify that the parsing of expression failed

**Expected behavior**
A unary-tests expression with a `some` or `every` expression can be parsed. 

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

saig0:
## Description

* fix parser failure that prevented the usage of some/every in a unary-tests expression
* extend the reserved words by `some` and `every` to prevent that a variable can be named as it

## Related issues

closes #280

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -17,7 +17,6 @@
 package org.camunda.feel.impl.parser
 
 import fastparse.JavaWhitespace._
-import fastparse.Parsed.Success
 import fastparse._
 import org.camunda.feel._
 import org.camunda.feel.syntaxtree._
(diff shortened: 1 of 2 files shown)


korthout (approved):
LGTM 👍 

💭 Note, that adding `every` and `some` to the reserved words is a breaking change. If versioning is done via semver, please keep this in mind

saig0:
> Note, that adding every and some to the reserved words is a breaking change. If versioning is done via semver, please keep this in mind

Good point. In this context, I would argue that it is okay because the names of operators should not be used for variables.

PR #306, opened 2021-07-28
Allow unary-tests without braces

### LINKED ISSUE #109: Unary test comparison without braces cannot be evaluated
**Describe the bug**
When evaluating `8 > 7` as unary test, the following error is displayed:
`failed to parse expression '8 > 7': [1.3] failure: end of input expected`

**To Reproduce**
Evaluate the expression `8 > 7` as unary test

**Expected behavior**
* The expression can be evaluated
* No error message is shown

**Environment**
FEEL Scala 1.11.0-alpha1

**Hints**
The failing expression can be used as simple expression.

**Alternative Solution**
 The expression `(8 > 7)` can be evaluated successfully

**Is this a bug according to the DMN FEEL 1.2 specification?**
* Yes!
* The DMN 1.1 spec on pg. 107 ("10.3.1.2 Grammar rules") says: The structure of a comparison is: `expression , ( "=" | "!=" | "<" | "<=" | ">" | ">=" ) , expression`
  * "Language keywords are enclosed in double quotes in the grammar rules" cf. pg. 107

saig0:
## Description

* fix a limitation of the parser to write unary-tests expressions with a boolean comparison but without braces
* the problem was caused by the backtracking since the first literal was parsed by `simpleValue`
* help the backtracking by defining that `simplePositiveUnaryTest` is not followed by something else except another `simplePositiveUnaryTest`

## Related issues

closes #109

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -234,10 +234,13 @@ object FeelParser {
   }
 
   private def unaryTests[_: P]: P[Exp] = P(
-    ("not" ~ "(" ~ positiveUnaryTests ~ ")").map(Not) |
-      positiveUnaryTests |
-      P("-").map(_ => ConstBool(true))
+    acceptAnyInputValue |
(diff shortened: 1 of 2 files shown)


korthout (approved):
LGTM 👍

PR #317, opened 2021-08-12
Add a builtin extract function that returns pattern matching strings from a text

### LINKED ISSUE #315: Add a function that extracts text based on regex
We need to extract a list of references (contract/client/ticket ids) from a text input and output them as a stringified list or a separated string of values (since a DMN output var value can only be string if I'm not wrong).

The custom function could have the following signature:

extract("input_text", "regular_expression")
e.g. extract("my references are 1234 and 1256", "12*") with the stringified list: "[1234, 1256]"
pro: no separator required in the following task
con: requires a string-to-list conversion function not available in FEEL

or alternatively

extract("input_text", "regular_expression", "separator")
e.g. extract("my references are 1234 and 1256", "12*", ";") with the val-sep string: "1234;1256"
pro: the following task can do a FEEL split(dmn-val-sep-values, ";")
con: requires to know the convention for which separator (e.g. ";") to use

**Related pull request**
https://github.com/camunda/feel-scala/pull/269

vincentgiraud:
## Description
See: https://github.com/camunda/feel-scala/issues/315#issuecomment-895781784

The problems with the existing functions:

replace()
is complicated if you want to ignore other parts of the string (i.e. negative lookahead)
is tricky if you don't know the number of matches
spit() doesn't include the search string

## Related issue and pull request

closes #315


src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
```diff
@@ -172,6 +174,23 @@ object StringBuiltinFunctions {
...
+      case List(ValString(input), ValString(pattern)) => {
+        val p = Pattern.compile(pattern)
+        val m = p.matcher(input)
+        val r = ListBuffer[String]()
+        val numPattern = pattern.r
+        val matches = numPattern.findAllIn(input)
+        matches.foreach { m =>
+          r += m
+        }
+        ValList(r.map(ValString).toList)
```
saig0, line 189: I would prefer a functional style and immutable collections. 

```
val matches = numPattern.findAllIn(input).map(ValString).toList
ValList(matches)
```

src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
```diff
@@ -172,6 +174,23 @@ object StringBuiltinFunctions {
     }
   )
 
+  private def extractFunction = builtinFunction(
+    params = List("input", "pattern"),
+    invoke = {
+      case List(ValString(input), ValString(pattern)) => {
+        val p = Pattern.compile(pattern)
```
saig0, line 181: Please have a look at the latest changes on master. For the other functions, I wrapped the creation of the pattern into a `Try` block to avoid exceptions in case of invalid patterns.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinStringFunctionsTest.scala
```diff
@@ -129,4 +129,14 @@ class BuiltinStringFunctionsTest
              ValString(""))))
   }
 
+  "An extract() function" should "return a list of strings matching a pattern" in {
```
saig0, line 132: Please add an additional test for the case that the pattern is invalid. See the latest changes from master.

saig0 (changes requested):
@vincentgiraud thank you for the contribution :tada:

It looks good :+1: 
Please rebase with `master` and apply the minor suggestions.

saig0 (approved):
I merge the PR and add the missing parts :rocket:

PR #322, opened 2021-08-18
Access weekday property of local date-time value

### LINKED ISSUE #319: I can access the weekday as property of a local date-time value
**Describe the bug**
Currently, I can't access the `weekday` property of a (local) date-time value that has no timezone component. 

The property is available for date-time values with timezone and date values. The property was introduced by #194.

**To Reproduce**
1. Evaluate the expression 
```
date and time("2018-12-10T10:30:01Z").weekday
```
2. Verify that it returns `null` instead of `1` 

**Expected behavior**
I can access the `weekday` property also for date-time values without timezone.

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

saig0:
## Description

* the weekday property is available for date-time values without timezone

## Related issues

closes #319

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -254,6 +254,12 @@
               <differenceType>7004</differenceType>
               <method>* $anonfun*(*)</method>
             </ignored>
+            <ignored>
+              <!-- allow new methods in the interface-->
+              <className>org/camunda/feel/**</className>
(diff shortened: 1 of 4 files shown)


korthout (approved):
LGTM 🍰

Issue #289, opened 2021-06-24
Split function invocation can throw an error


saig0:
**Describe the bug**
If the `split()` function is invoked with an invalid delimiter then the engine throws an exception.

```
java.util.regex.PatternSyntaxException: Unexpected internal error near index 1
\
	at java.util.regex.Pattern.error(Unknown Source) ~[?:?]
	at java.util.regex.Pattern.compile(Unknown Source) ~[?:?]
	at java.util.regex.Pattern.<init>(Unknown Source) ~[?:?]
	at java.util.regex.Pattern.compile(Unknown Source) ~[?:?]
	at org.camunda.feel.impl.builtin.StringBuiltinFunctions$$anonfun$splitFunction$1.applyOrElse(StringBuiltinFunctions.scala:179) ~[feel-engine-1.12.4.jar:1.12.4]
	at org.camunda.feel.impl.builtin.StringBuiltinFunctions$$anonfun$splitFunction$1.applyOrElse(StringBuiltinFunctions.scala:177) ~[feel-engine-1.12.4.jar:1.12.4]
	at scala.PartialFunction$OrElse.apply(PartialFunction.scala:241) ~[scala-library-2.13.4.jar:?]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.invokeFunction(FeelInterpreter.scala:775) ~[feel-engine-1.12.4.jar:1.12.4]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.$anonfun$eval$40(FeelInterpreter.scala:189) ~[feel-engine-1.12.4.jar:1.12.4]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.withFunction(FeelInterpreter.scala:742) ~[feel-engine-1.12.4.jar:1.12.4]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.eval(FeelInterpreter.scala:189) ~[feel-engine-1.12.4.jar:1.12.4]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.$anonfun$withCartesianProduct$1(FeelInterpreter.scala:823) ~[feel-engine-1.12.4.jar:1.12.4]
	at scala.collection.immutable.List.map(List.scala:246) ~[scala-library-2.13.4.jar:?]
	at org.camunda.feel.impl.interpreter.FeelInterpreter.withCartesianProduct(FeelInterpreter.scala:823) ~[feel-engine-1.12.4.jar:1.12.4]
[...]

PR #323, opened 2021-08-18
Failsafe parse/eval API calls

saig0:
## Description

* avoid unexpected exceptions when calling the parse/eval API
* return always an `either` instead of letting the exception bubble up

## Related issues

avoid exceptions like #289


src/main/scala/org/camunda/feel/FeelEngine.scala
```diff
@@ -196,7 +202,12 @@ class FeelEngine(
   }
 
   def eval(exp: ParsedExpression, context: Context): EvalExpressionResult =
-    validate(exp).flatMap(_ => eval(exp, rootContext + context))
+    Try {
+      validate(exp).flatMap(_ => eval(exp, rootContext + context))
+    }.recover(failure =>
+        Left(
+          Failure(s"failed to evaluate expression '${exp.text}' : $failure")))
+      .get
```
korthout, line 210: Should like 214 also be wrapped inside a Try?

korthout (approved):
Nice improvement!

LGTM, just 1 question but if the answer is yes it can also be done in a separate PR.

src/main/scala/org/camunda/feel/FeelEngine.scala
saig0: It is possible but not necessary. 

The method

```
def eval(exp: ParsedExpression, variables: Map[String, Any] = Map()): EvalExpressionResult
```

delegates to

```
def eval(exp: ParsedExpression, context: Context): EvalExpressionResult
```

So, the evaluation is already wrapped in a `Try`.

PR #330, opened 2021-09-01
FEEL parser overhaul :wrench: 

### LINKED ISSUE #312: Fail to parse instance of with a function definition
**Describe the bug**
The engine fails to parse an expression with `instance of` if the value is a function definition.

**To Reproduce**
1. Evaluate the expression 
```(function() "foo") instance of Any```
2. Verify that it fails to parse the expression

**Expected behavior**
The engine parses an expression with `instance of` and a function definition.

```
(function() "foo") instance of Any
// true
```

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

saig0:
## Description

* rewrite the whole parser (to get rid of some old parts) 
* simplify the parsers by improving the hierarchy of the expressions
* removed unused parsers

**Hints**
I recommend looking just at the new parser. The comparison of the changes is not so useful because the structure of the parser changed too much. 

## Related issues

closes #312

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -39,526 +39,508 @@ object FeelParser {
   def parseUnaryTests(expression: String): Parsed[Exp] =
     parse(expression, fullUnaryExpression(_))
 
-  private def fullExpression[_: P] = P(Start ~ expression ~ End)
+  // --------------- entry parsers ---------------
 
-  private def fullUnaryExpression[_: P] = P(Start ~ unaryTests ~ End)
+  private def fullExpression[_: P]: P[Exp] = P(Start ~ expression ~ End)
 
-  private def reservedWord[_: P] =
+  private def fullUnaryExpression[_: P]: P[Exp] = P(Start ~ unaryTests ~ End)
+
+  // --------------- common parsers ---------------
+
+  // language key words that can't be used as variable names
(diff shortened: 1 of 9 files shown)


korthout (approved):
👏 🏆 🏅 Great improvement!

First of all, the parser has become much more readable/understandable. In particular the overall structure of the parser has improved greatly by the grouping, renaming and the added comments. I really love it.

What I also love about this PR is that you were able to leave all existing test cases unchanged. In face, you only added new ones to further test everything.

Like you already mentioned, the changes are so large, that it was hard to track changes. In some cases, I still felt the need to compare the implementations of specific parsers because I did not fully understand how they did what their name implied. For example, why can an `identifier` start with a `reservedWord`, but not contain one in the middle? I was also unable to find a test for it. Comparing the new code with the old, showed that the implementation actually hadn't changed, so we can leave it out of scope. But, I wanted to have mentioned it in the review.

saig0:
@korthout I'm happy to hear that you like the changes and see a benefit. Thank you for the nice words ❤️ 

Regarding reserved words in identifiers, reserved words should also be allowed in between or at the end of an identifier. (And yes, we should add tests)

https://github.com/camunda/feel-scala/blob/fc2971280691555785270fe93f6a840f3e3da8da/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala#L62

The parser uses a short notation that should avoid that an identifier is just a reserved word.

PR #336, opened 2021-10-01
bug(builtin-function): add 'm' parameter support to get entries

### LINKED ISSUE #320: get entries() expects a different argument name
**Describe the bug**
In the DMN 1.3 spec, the function `get entries()` defines the argument: `m` .

But in the engine, the function expects the argument: `context`. 
Background: the function was introduces in the engine before it was introduced in DMN 1.3. 

**To Reproduce**
1. Evaluate the expression 
```
get entries(m:{a: "foo"})
```
2. Verify that the expression evaluates to `null` instead of `"foo"` 

**Expected behavior**
The `get entries()` function can be invoked with the named parameter: `m`.

For backward compatibility, the function should still support the argument: `context`.

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

P3trur0:
## Description

* Adds the support to `m` argument  name to `get entries` builtin function. 

## Related issues

closes #320

### DIFF
--- a/docs/versioned_docs/version-1.14/reference/builtin-functions/feel-built-in-functions-context.md
+++ b/docs/versioned_docs/version-1.14/reference/builtin-functions/feel-built-in-functions-context.md
@@ -22,11 +22,11 @@ get value({foo: 123}, "foo")
 Returns the entries of the context as list of key-value-pairs.
 
 * parameters:
-  * `context`: context
+  * `m`: context
 * result: list of context which contains two entries for "key" and "value"
 
 ```js
-get entries({foo: 123})
+get entries(m:{foo: 123})
 // [{key: "foo", value: 123}]
 ```
 
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -254,8 +254,6 @@ class BuiltinContextFunctionsTest
       ValBoolean(true))
     eval(""" context(get entries({a:1,b:2})) = {a:1, b:2} """) should be(
       ValBoolean(true))
-    eval(""" context(get entries(m:{a:1,b:2})) = {a:1, b:2} """) should be(
-      ValBoolean(true))
     eval(""" context(get entries({a:1,b:2})[key="a"]) = {a:1} """) should be(
       ValBoolean(true))
   }
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -51,17 +51,12 @@ class BuiltinContextFunctionsTest
   "A get entries function" should "return all entries when invoked with 'm' argument" in {
 
     val list = eval(""" get entries(m:{foo: 123}) """)
-    list shouldBe a[ValList]
-
-    val items = list.asInstanceOf[ValList].items
-    items should have size 1
-    val context = items(0)
-    context
-      .asInstanceOf[ValContext]
-      .context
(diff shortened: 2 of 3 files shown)


saig0:
I'll review it soon (~ beginning of next week).

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -48,6 +48,22 @@ class BuiltinContextFunctionsTest
       Map("key" -> ValString("foo"), "value" -> ValNumber(123)))
   }
 
+  "A get entries function" should "return all entries when invoked with 'm' argument" in {
```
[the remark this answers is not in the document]
saig0: :x: We should use the old parameter name in the previous test case. So, we can verify that the previous behavior still works.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -243,6 +259,8 @@ class BuiltinContextFunctionsTest
       ValBoolean(true))
     eval(""" context(get entries({a:1,b:2})) = {a:1, b:2} """) should be(
       ValBoolean(true))
+    eval(""" context(get entries(m:{a:1,b:2})) = {a:1, b:2} """) should be(
+      ValBoolean(true))
```
saig0, line 263: :wrench: We don't need the additional case here. 

This test case verifies the behavior of the `context()` function in combination with `get entries()`. The argument name is not relevant in this test case.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -48,6 +48,22 @@ class BuiltinContextFunctionsTest
...
+
+    val items = list.asInstanceOf[ValList].items
+    items should have size 1
+    val context = items(0)
+    context
+      .asInstanceOf[ValContext]
+      .context
+      .variableProvider
+      .getVariables should be(
+      Map("key" -> ValString("foo"), "value" -> ValNumber(123)))
```
saig0, line 64: :wrench: We could shrink the verification a bit.

```suggestion
    list match {
      case ValList(List(ValContext(context))) =>
        context.variableProvider.getVariables should be(
          Map("key" -> ValString("foo"), "value" -> ValNumber(123)))
      case other => fail(s"Expected list with one context but found '$other'")
    }
```

docs/versioned_docs/version-1.14/reference/builtin-functions/feel-built-in-functions-context.md
```diff
@@ -22,11 +22,11 @@ get value({foo: 123}, "foo")
 Returns the entries of the context as list of key-value-pairs.
 
 * parameters:
-  * `context`: context
+  * `m`: context
 * result: list of context which contains two entries for "key" and "value"
 
 ```js
-get entries({foo: 123})
+get entries(m:{foo: 123})
```
saig0, line 29: :+1: Thank you for updating the docs!

saig0 (changes requested):
@P3trur0 your PR looks good :+1: 

I've just some minor suggestions. Please have a look.

docs/versioned_docs/version-1.14/reference/builtin-functions/feel-built-in-functions-context.md
P3trur0: u're welcome: I thought it was better to update them for the sake of completeness 😄

P3trur0:
@saig0 please find PR updated as per your suggestions!

saig0:
@P3trur0 thanks :+1: I merged your PR :tada:

PR #337, opened 2021-10-01
bug(builtin-function): add 'm' and 'key' parameters support to get value

### LINKED ISSUE #313: get value() expect different argument names
**Describe the bug**
In the DMN 1.3 spec, the function `get value()` defines the arguments: `m` and `key`.

But in the engine, the function expects the arguments: `context` and `key`. 
Background: the function was introduces in the engine before it was introduced in DMN 1.3. 

**To Reproduce**
1. Evaluate the expression 
```
get value(key:"a", m:{a: "foo"})
```
2. Verify that the expression evaluates to `null` instead of `"foo"` 

**Expected behavior**
The `get value()` function can be invoked with the named parameters: `m` and `key`.

For backward compatibility, the function should still support the arguments: `context` and `key`.

**Environment**
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: [0.x]

P3trur0:
## Description

* Adds `m` and `key` parameters support to `get value`

## Related issues

closes #313

### DIFF
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -57,7 +57,7 @@ class BuiltinContextFunctionsTest
     eval(""" get value({foo: 123}, "foo") """) should be(ValNumber(123))
   }
 
-  "A get value function" should "return the value when arguments are named 'm' and 'key'" in {
+  it should "return the value when arguments are named 'm' and 'key'" in {
     eval(""" get value(m:{foo: 123}, key:"foo") """) should be(ValNumber(123))
   }
 
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -54,13 +54,16 @@ class BuiltinContextFunctionsTest
   }
 
   "A get value function" should "return the value" in {
-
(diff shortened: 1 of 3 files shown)


saig0:
I'll review your PR soon (~ beginning of next week).

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -54,13 +54,16 @@ class BuiltinContextFunctionsTest
   }
 
   "A get value function" should "return the value" in {
-
     eval(""" get value({foo: 123}, "foo") """) should be(ValNumber(123))
   }
 
-  it should "return null if not contains" in {
+  "A get value function" should "return the value when arguments are named 'm' and 'key'" in {
```
saig0, line 60: :x: We should group test cases by using it. The test case will be in the same group as the previous test case (i.e. "A get value function").

```suggestion
  it should "return the value when arguments are named 'm' and 'key'" in {
```

saig0: :x: We should use the old parameter name in the previous test case. So, we can verify that the previous behavior still works.

saig0 (changes requested):
@P3trur0 your PR looks good :+1: 

I've just a few minor suggestions. Please have a look.

P3trur0:
@saig0 here you go, suggestions merged.

saig0 (approved):
@P3trur0 thanks :+1: Your PR is ready to merge :tada:

PR #340, opened 2021-10-05
bug(builtin-function): fix get entries to return in the same order

### LINKED ISSUE #161: get entries() function returns the entries in a different order
**Describe the bug**
The function `get entries()` may return the entries in a different order than they are defined in the context.

**To Reproduce**
Expression `get entries({a: "foo", b: "bar"})`
 returns 
`[{value=bar, key=b}, {value=foo, key=a}]` 
but should be 
`[{value=foo, key=a}, {value=bar, key=b}]`

**Expected behavior**
The function `get entries()` returns the entries in the same order as they are defined in the context.

**Environment**
* FEEL engine version: 1.11.2

P3trur0:
## Description

* this reverses the order of variables provided by `VariableProvider` since the AST interpreter swaps the original order of the provided arguments.  

Probably this bug fix may look naive, but it does the trick. I picked this approach for two main reasons:

1 - the assumption that input lists are not very long
2 - my poor knowledge of the interpreter codebase.

## Related issues

closes #161


saig0:
I think that the PR goes in the right direction. The current chaining in the `EvalContext` is too complicated and not necessary. We should simplify the logic and use a plain map (or sorted map) instead of chaining the contexts with every new entry. 

Since this is a bigger change and a slightly different focus, I'm closing the PR in favor of a new PR that changes the behavior of `EvalContext` first.

@P3trur0 thank you again for your effort :cake: It helped me to root cause the issue and find a suitable approach.

PR #354, opened 2021-10-28
feat(builtinFunction): Range - Add last few range functions

### LINKED ISSUE #178: Add new built-in range functions
The built-in range functions below were added to the [DMN 1.3 standard](https://www.omg.org/spec/DMN/1.3/PDF) (see § 10.3.4.7 Range Functions). These functions would be helpful when writing business rules with FEEL:

- before
- after
- meets
- met by
- overlaps before
- overlaps after
- finishes
- finished by
- includes
- during
- starts
- started by
- coincides

> The following set of functions establish relationships between single scalar values and ranges of such values. All
functions in this list take two arguments and return True if the relationship between the argument holds, or False
otherwise.

image

image

image

image

image

image

image

image

image

kaaquist:
## Description

* added "finished by" plus tests
* added "includes" plus tests
* added "during" plus tests
* added "starts" plus tests
* added "started by" plus tests
* added "coincides" plus tests

Documentation on Range functions in general will be added in a new PR.  

## Related issues
Last part of #178 
 

closes #178


saig0 (approved):
:1st_place_medal:

saig0:
@kaaquist I merged your PR :rocket: I did some small changes to support not only numbers but also the other data types :grin:

PR #358, opened 2021-11-03
Avoid StackOverflowError with huge lists

### LINKED ISSUE #350: Filter on large list causes StackOverflowError
**Describe the bug**
A filter operation on a large list (more than 1000 items) causes a `StackOverflowError`.

Screenshot from 2021-10-14 06-45-29

**To Reproduce**
1. Evaluate the following expression
```
list[even(item)]
```
  * with a variable `list` that contains more than 1000 items (e.g. `List.range(1,1500)`)
2. Verify that a `StackOverflowError` is thrown

**Expected behavior**
I can evaluate a filter operation on large lists (without "any" limitation regarding the size of the list).

**Environment**
* FEEL engine version: `1.13.3`
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: `1.1.2` https://github.com/camunda-cloud/zeebe/issues/7758

saig0:
## Description

* avoid StackOverflowError if one of the following operations is applied to a huge list (e.g. 10.000 items)
  * `filter` 
  * `some`
  * `every`
  * `all()`
  * `any()`
  * `flatten()`

## Related issues

closes #350

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
@@ -215,18 +215,13 @@ object ListBuiltinFunctions {
       case List(ValList(list)) => all(list)
     }, hasVarArgs = true)
 
-  private def all(xs: List[Val]): Val = xs match {
-    case Nil => ValBoolean(true)
-    case x :: xs =>
-      x match {
-        case ValBoolean(false) => ValBoolean(false)
-        case ValBoolean(true)  => all(xs)
-        case other =>
-          all(xs) match {
-            case ValBoolean(false) => ValBoolean(false)
-            case _                 => ValNull
-          }
-      }
+  private def all(items: List[Val]): Val = {
+    items.foldLeft[Val](ValBoolean(true)) {
+      case (ValBoolean(false), _)               => ValBoolean(false)
+      case (ValBoolean(true), item: ValBoolean) => item
+      case (ValNull, ValBoolean(false))         => ValBoolean(false)
+      case (_, _)                               => ValNull
+    }
   }
 
   private def orFunction =
@@ -234,18 +229,13 @@ object ListBuiltinFunctions {
       case List(ValList(list)) => atLeastOne(list)
     }, hasVarArgs = true)
 
-  private def atLeastOne(xs: List[Val]): Val = xs match {
-    case Nil => ValBoolean(false)
-    case x :: xs =>
-      x match {
-        case ValBoolean(true)  => ValBoolean(true)
-        case ValBoolean(false) => atLeastOne(xs)
-        case other =>
-          atLeastOne(xs) match {
-            case ValBoolean(true) => ValBoolean(true)
-            case _                => ValNull
-          }
-      }
+  private def atLeastOne(items: List[Val]): Val = {
+    items.foldLeft[Val](ValBoolean(false)) {
+      case (ValBoolean(true), _)                 => ValBoolean(true)
+      case (ValBoolean(false), item: ValBoolean) => item
+      case (ValNull, ValBoolean(true))           => ValBoolean(true)
+      case (_, _)                                => ValNull
+    }
   }
 
   private def sublistFunction =
@@ -362,10 +352,11 @@ object ListBuiltinFunctions {
       case List(ValList(list)) => ValList(flatten(list))
     })
 
-  private def flatten(list: List[Val]): List[Val] = list match {
-    case Nil              => Nil
-    case ValList(l) :: xs => flatten(l) ++ flatten(xs)
-    case x :: xs          => x :: flatten(xs)
+  private def flatten(list: List[Val]): List[Val] = {
+    list.flatten {
+      case ValList(items) => flatten(items)
+      case item           => List(item)
+    }
   }
 
   private def sortFunction = builtinFunction(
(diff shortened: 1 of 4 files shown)


src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -538,37 +538,32 @@ class FeelInterpreter {
...
+    items.foldLeft(f(false)) {
+      case (ValBoolean(true), _) => f(true)
+      case (ValNull, item) =>
+        item() match {
+          case ValBoolean(true) => f(true)
+          case _                => ValNull
         }
+      case (_, item) => withBooleanOrNull(item(), f)
     }
+  }
```
korthout, line 551: ❓ I don't understand the method's signature. Can you explain why `f` is a function that takes a `Boolean` and returns a `Val`? I'd expected a `Predicate<Val>` (or similar) for f.

src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
```diff
@@ -215,37 +215,27 @@ object ListBuiltinFunctions {
...
-          all(xs) match {
-            case ValBoolean(false) => ValBoolean(false)
-            case _                 => ValNull
-          }
-      }
+  private def all(items: List[Val]): Val = {
+    items.foldLeft[Val](ValBoolean(true)) {
+      case (ValBoolean(false), _)               => ValBoolean(false)
+      case (ValBoolean(true), item: ValBoolean) => item
+      case (ValNull, ValBoolean(false))         => ValBoolean(false)
```
korthout, line 222: ❌ I think L222 might be wrong, as it could turn a null case back to false, e.g. `all([true, null, false]) => false, but should be null`. Perhaps I misunderstand this though.

korthout (commented):
Nice work 🚀  But it's also good that I'm gonna get that scala training because I have some questions 😆 

Please have a look

src/test/scala/org/camunda/feel/impl/builtin/BuiltinListFunctionsTest.scala
```diff
@@ -207,6 +219,18 @@ class BuiltinListFunctionsTest
...
+
+    eval("any(xs)", Map("xs" -> hugeList)) should be (ValBoolean(false))
+  }
+
+  it should "return null if items are not boolean values (huge list)" in {
+    val hugeList = (1 to 10_000).toList
+
+    eval("any(xs)", Map("xs" -> hugeList)) should be (ValNull)
+  }
+
```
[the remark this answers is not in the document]
saig0: Nope :sweat_smile: These test cases are similar but using the function `any()` instead of `all()` :mag:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -945,16 +945,16 @@ class FeelInterpreter {
...
+  private def filterList(list: List[Val], filter: Val => Val): Val = {
+    val conditionNotFulfilled = ValString("_")
+
+    mapEither[Val, Val](list,
+      item => withBoolean(filter(item), {
+        case true => item
+        case false => conditionNotFulfilled
+      }).toEither,
+      items => ValList(items.filterNot(_ == conditionNotFulfilled)))
+  }
```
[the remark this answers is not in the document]
saig0: It would be nice to use `filter()` but we need to handle also the failure cases. If a filter doesn't evaluate to a boolean then we want to report the failure. 

I'm not sure how to combine this failure handling with a `filter()`. I'm open to suggestions.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: The parameter `f` is the constructor for the return value `ValBoolean`. 

Currently, this construct is used in many places in the interpreter. I plan to simplify it but don't want to do it now :sweat_smile:

src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
saig0: This is by intention :sweat_smile: 

In FEEL, if one value is `false` then the result is `false`. The value of `null` (or any other non-boolean value) can't change the result. Only if all other values are `true` then the value of `null` matters because it could change the result to `false`.

https://camunda.github.io/feel-scala/docs/reference/language-guide/feel-boolean-expressions#conjunction--and

src/test/scala/org/camunda/feel/impl/builtin/BuiltinListFunctionsTest.scala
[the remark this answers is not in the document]
korthout: 🤦 Oops 😄

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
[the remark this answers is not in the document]
korthout: Ah I understand it now. Evaluation errors result in `ValError` which is also a `Val` and turns into a left with the `.toEither`. It makes sense to me now.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
korthout: I understand it now, thanks 👍

src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
korthout: Ah that's different than I expected. Interesting

korthout (approved):
Thanks for changes 🚀 

LGTM 👍

saig0:
@korthout thank you for your fast review :heart:

PR #381, opened 2021-12-30
chore(interpreter): add a test for whitespaced named parameters

### LINKED ISSUE #311: I can invoke a function with a named parameter that contains whitespaces
I can invoke a function with a named parameter. 

If the named parameter contains whitespace then the engine fails to parse the expression. 

A named parameter of a function invocation can contain whitespaces.

**Related issues**

* Camunda BPM: 
* Zeebe broker:

P3trur0:
## Description

* this adds a unit test for #311.  

Please, mind that I build the test based on [this](https://github.com/camunda/tck/issues/7). There, the function used is:

`number(from: "1.000.000,01", decimal sep:",", grouping sep:".")`

while in the codebase here the expected parameters names are:

- `from`, `decimal separator`, `grouping separator`

## Related issues

This tests the behavior expected by #311.

closes #311

### DIFF
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
@@ -179,4 +179,8 @@ class InterpreterFunctionTest
     eval("round(3.2)", functions = functions) should be(ValNumber(3))
   }
 
+  it should "be properly evaluated when parameters containing whitespaces" in {
+    eval("""number(from: "1.000.000,01", decimal separator:",", grouping separator:".")""") should be(ValNumber(1_000_000.01))
+  }
+
 }
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
@@ -138,6 +138,18 @@ class InterpreterFunctionTest
     eval(""" index of([1,2,3,2],2)[1]  """) should be(ValNumber(2))
   }
 
+  it should "be properly evaluated when parameters contain whitespaces" in {
+    eval("""number(from: "1.000.000,01", decimal separator:",", grouping separator:".")""") should be(ValNumber(1_000_000.01))
+  }
+
+  it should "be invoked with one named parameter containing whitespaces" in {
+    val functions =
+      Map("f" -> eval("""function(test name) `test name` + 1""").asInstanceOf[ValFunction])
+
+    eval("f(test name:1)", functions = functions) should be(ValNumber(2))
+    eval("f(test name:2)", functions = functions) should be(ValNumber(3))
+  }
+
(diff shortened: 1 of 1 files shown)


src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
```diff
@@ -179,4 +179,8 @@ class InterpreterFunctionTest
     eval("round(3.2)", functions = functions) should be(ValNumber(3))
   }
 
+  it should "be properly evaluated when parameters contain whitespaces" in {
```
saig0, line 182: :x: The test case should be grouped under `"A function definition"`.

Currently, it is grouped under `"An external java function definition"` that is the last test group in this class. We should move the test case up, for example, after `it should "be invoked and followed by a filter"`. It is not related to external functions.

image

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
```diff
@@ -179,4 +179,8 @@ class InterpreterFunctionTest
...
+    eval("""number(from: "1.000.000,01", decimal separator:",", grouping separator:".")""") should be(ValNumber(1_000_000.01))
```
saig0, line 183: :wrench: We could create another test case for a user-defined function.

Like [here](https://github.com/camunda/feel-scala/blob/master/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala#L48-L55) but with a function parameter that has whitespace. Take the function definition as an argument and invoke the function in the test case.

saig0 (changes requested):
@P3trur0 thank you for providing the test case :+1: 

I have two minor comments. Please have a look :cookie:

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
P3trur0: my bad, gonna fix it!

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterFunctionTest.scala
P3trur0: Ok! Do you mean anything like:

```scala
    val functions =
      Map("f" -> eval("""function(test name) test name + 1""").asInstanceOf[ValFunction])
```

?

saig0: Yes :+1: But I guess that you need to escape the name in the function body. Names with whitespaces are not fully supported :sweat_smile: 

```
    val functions =
      Map("f" -> eval("""function(test name) `test name` + 1""").asInstanceOf[ValFunction])

    eval("f(test name: 1)", functions = functions) should be(ValNumber(2))
```

saig0 (approved):
@P3trur0 nice :+1:

PR #383, opened 2021-12-31
handle null values in max and min list functions

### LINKED ISSUE #382: list function max not handling null values
**Describe the bug**
When using the list function max, if one of the values is a ValNull the function will return a null value as it is deemed as not comparable (I assume this could also be a bug for other list functions as well)

**To Reproduce**
max(null, 0.005, 0.005)

has error of:
_org.camunda.feel.FeelEngine - ValList(List(ValNull, ValNumber(0.5), ValNumber(0.5))) is not comparable_
and it returns: _ValNull_

**Expected behavior**
examples
max(null, 0.005, 0.005) = 0.005
max(null, 0.002, 0.005) = 0.005

hawksdoves:
## Description
 isComparable for Lists now handles ValNull as does the list functions min, max. The behaviour is such min and max will just ignore null values.

## Related issues
382

closes #382

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
@@ -64,7 +64,7 @@ object ListBuiltinFunctions {
       case List(l @ ValList(list)) =>
         list match {
           case Nil                   => ValNull
-          case _ if (l.isComparable) => list.min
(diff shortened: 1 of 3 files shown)


saig0:
@hawksdoves thank you for the contribution :tada:

Before I will check the code, we need to clarify the issue #382 first. Please have a look at my comment: https://github.com/camunda/feel-scala/issues/382#issuecomment-1003939937

saig0:
Closing the PR for now until the requirements are clarified.

PR #385, opened 2022-01-03
fix(engine): list projection with variable

### LINKED ISSUE #384: Fail to extract a property of a list of contexts (list projection)
**Describe the bug**
The evaluation of a `path` expression fails if the value is a list of contexts (i.e. a list projection). The expression worked in previous versions of the FEEL engine.

image

Note that the `path` expression works if the value is a list literal instead of a variable.

**To Reproduce**
1. Evaluate the expression `x.y`
  * with variable `x` = `[{y:1},{y:2}]`

**Expected behavior**
I can evaluate a `path` expression if the variable is a list of contexts. Same as for list literals.

**Environment**
* FEEL engine version: `1.14.0`
* Affects: 
  * Camunda BPM: [7.x] 
  * Zeebe broker: `1.3.0` https://github.com/camunda-cloud/zeebe/pull/8494

saig0:
## Description

* fix path expression for list of contexts (i.e. list projection) with variable reference
  * reuse the existing logic of `path` expressions to avoid duplication and a difference in the behavior 
* it worked previously because it was parsed as a `path` expression instead of a `ref` expression

## Related issues

closes #384

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -978,17 +978,9 @@ class FeelInterpreter {
     names match {
       case Nil => x
       case n :: ns =>
-        withContext(
-          x, {
-            case ctx: ValContext =>
-              EvalContext
-                .wrap(ctx.context)(context.valueMapper)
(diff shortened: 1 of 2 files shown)


korthout (approved):
LGTM 🍄

PR #388, opened 2022-01-11
fix(parser) managing arbitrary names for a context key

### LINKED ISSUE #310: I can use arbitrary names for a context key
I can define a context literal with arbitrary names for the context keys. For example, using a context key with whitespace or symbol. 

A context key can be an arbitrary name, including whitespace and symbols.

```
{foo bar: "foo"}

{foo+bar: "foo"}
```

**Related issues**

* Camunda BPM: 
* Zeebe broker:

P3trur0:
## Description

* introduces new parsing options for context key names

As you may notice, I slightly changed also the behavior of `private def identifierWithWhitespaces[_: P]: P[String]` since in previous version it supported only names with a single whitespace.

## Related issues

closes #310

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -420,10 +420,7 @@ object FeelParser {
   private def context[_: P]: P[Exp] =
     P(
       "{" ~ contextEntry.rep(0, sep = ",") ~ "}"
-    ).map{
-      entries =>
-        ConstContext(entries.toList)
-    }
+    ).map(entries => ConstContext(entries.toList))
 
   private def contextEntry[_: P]: P[(String, Exp)] = P(
     (contextAnySymbol | identifierWithWhitespaces | name | stringWithQuotes) ~ ":" ~ expression
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -169,7 +169,7 @@ object FeelParser {
   // use only if the identifier is followed by a predefined character (e.g. `(` or `:`)
   private def identifierWithWhitespaces[_: P]: P[String] =
     P(
-      identifier ~~ (" " ~~ identifier).repX(1)
+      identifier ~~ (" ".repX(1) ~~ identifier).repX(1)
     ).!
 
   private def name[_: P]: P[String] = P(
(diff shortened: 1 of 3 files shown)


saig0:
I'll have a look soon :eyes: 

I did expect this feature :sweat_smile: but one step forwards the DMN spec :rocket:

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -423,9 +423,13 @@ object FeelParser {
     ).map(entries => ConstContext(entries.toList))
 
   private def contextEntry[_: P]: P[(String, Exp)] = P(
-    (name | stringWithQuotes) ~ ":" ~ expression
+    (contextAnySymbol | identifierWithWhitespaces | name | stringWithQuotes) ~ ":" ~ expression
   )
 
+  private def contextAnySymbol[_: P]: P[String] = {
+    P((!"\"" ~ !"{" ~ !"}" ~ !":" ~ !"," ~ AnyChar ~ !"{").rep(1).!)
+  }
```
saig0, line 431: :wrench: We could extract the reserved symbols in a method. This would make it a bit easier to read.

```
private def reservedSymbol[_: P]: P[String] = P(
    CharIn("\"" , "{" , "}" , ":" , ",", "[", "]", "`")
  ).!

private def contextKeyAnySymbol[_: P]: P[String] = P(
    (!reservedSymbol ~ AnyChar).rep(1)
  ).!
```

saig0 (changes requested):
@P3trur0 the PR looks good :tada: 
I like your changes :+1: 

I have one suggestion for the parser. Please have a look :cookie:

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
P3trur0: Cool, it looks way better!

P3trur0: Method added. As a matter of clean code principles, I swapped the order of methods, namely:

```
private def contextKeyAnySymbol[_: P]: P[String] = P(
    (!reservedSymbol ~ AnyChar).rep(1)
  ).!
  
  private def reservedSymbol[_: P]: P[String] = P(
    CharIn("\"" , "{" , "}" , ":" , ",", "[", "]", "`")
  ).!
```

saig0 (approved):
@P3trur0 thanks :+1: The PR is ready to merge :rocket:

PR #395, opened 2022-01-27
fix(engine): add `satisfies`, `return`, `then`, `else` as reserved keywords

### LINKED ISSUE #394: Evaluation of expression `keyword myVariable keyword myFunction()` fails
**Describe the bug**

The evaluation of an expression structured as `myVariable keyword myFunction()` fails with a `ValError(no function found with name 'myVariable keyword myFunction' and 1 parameters)`.

**To Reproduce**

Evaluate the following expression to reproduce the problem:

```scala
eval("if 7 > var then flatten(xs) else []",  Map("xs" -> List(1, 2), "var" -> 3))
```

**Expected behavior**

The expression has the following return value:

```scala
ValList(List(ValNumber(1), ValNumber(2)))
```

**Environment**
* FEEL engine version: 1.14.1
* Affects: 
  * Camunda Automation Platform 7: [7.x] https://jira.camunda.com/browse/CAM-14316
  * Zeebe broker: [0.x]

tasso94:
## Description

Add `satisfies`, `return`, `then`, `else` as reserved keywords.

## Related issues

closes #394

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -20,5 +20,5 @@ A clear and concise description of what you expected to happen.
 **Environment**
 * FEEL engine version: [1.x.y]
 * Affects: 
-  * Camunda BPM: [7.x] <!-- link the issue: https://jira.camunda.com/browse/CAM- -->
+  * Camunda Automation Platform 7: [7.x] <!-- link the issue: https://jira.camunda.com/browse/CAM- -->
   * Zeebe broker: [0.x] <!-- link the issue: https://github.com/zeebe-io/zeebe/issues# -->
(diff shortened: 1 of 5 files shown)


saig0 (approved):
@tasso94 looks good :+1:

tasso94:
Hi @saig0,

Thank you for your review.

I found out that `and` as well as `or` currently suffer under the same problem, and we should probably fix it as well. Unfortunately, adding these keywords to the list of reserved words leads to many failing test cases.

Best,
Tassilo

saig0:
@tasso94 nice finding 👍 

Let me know if I should look into it or if we should look into it together 👀

saig0:
@tasso94 I created a new issue for expressions with `and` / `or`: #397. We see this issue now also in Zeebe :sweat_smile: I may take a look this week.

This PR seems to be ready. Can we merge it and backport it to `1.3.x`?

saig0:
@tasso94 true. I wanted to say that we should backport to branch `1.14.x` :sweat_smile:

tasso94:
Got it! Will do 🙂👍

PR #399, opened 2022-02-09
Expression with conjunction/disjunction (and/or) fails if it contains parentheses

### LINKED ISSUE #397: Expression with conjunction/disjunction (and/or) fails if it contains parentheses
**Describe the bug**
If an expression contains a conjunction/disjunction (and/or) and the second argument is wrapped into parentheses then the expression can't be evaluated. The engine parses the conjunction/disjunction as a function call.

image

**To Reproduce**
1. Parse the expression `x and (y)` 
2. Evaluate the expression with `x` and `y` as boolean values
3. Verify that the evaluation fails with the message `failed to evaluate expression ' x and (y) ': no function found with name 'x and' and 1 parameters`

**Expected behavior**
The expression is parsed as conjunction/disjunction, instead of a function call, and it can be evaluated. 

**Environment**
* FEEL engine version: `1.14.1`
* Affects: 
  * Camunda Automation Platform 7: https://jira.camunda.com/browse/CAM-14344
  * Zeebe broker: `1.3.2` https://github.com/camunda-cloud/zeebe/issues/8740

saig0:
## Description

* add the reserved keywords: "and", "or" to avoid that a conjunction/disjunction is parsed as a function name
* change the parsing behavior to avoid that a function name contains a reserved keyword like "and" / "or"
* list all built-in function names with keywords explicitly because they don't match the regular function name anymore

## Related issues

closes #397

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -147,7 +147,9 @@ object FeelParser {
         "return",
         "then",
         "else",
-        "satisfies"
+        "satisfies",
+        "and",
+        "or"
       )
     ).!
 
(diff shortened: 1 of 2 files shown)


tasso94 (approved):
Looks good to me! 🙂 👍

PR #417, opened 2022-04-25
Filter list with custom function

### LINKED ISSUE #359: Custom function invoked with ValError(no variable found for name 'item') when used in list filter condition
**Describe the bug**
We have a custom function custom_even and trying to use it in the following decision:
```xml
<decision id="item" name="item not resolved demo">
    <variable id="InformationItem_05x2m7o" name="result" typeRef="string" />
    <literalExpression id="LiteralExpression_0g7w721" expressionLanguage="feel">
      <text>[1,2,3,4,5][custom_even(item)]</text>
    </literalExpression>
  </decision>
```
The function is invoked 6 times: first invocation is with the **ValError(no variable found for name 'item')** argument, 5 invocations with the actual list elements.

**To Reproduce**
This code snippet can be used to reproduce the issue:
```java
public class DmnEngineTest {
    static final String DMN = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
            "<definitions xmlns=\"https://www.omg.org/spec/DMN/20191111/MODEL/\" xmlns:dmndi=\"https://www.omg.org/spec/DMN/20191111/DMNDI/\" xmlns:dc=\"http://www.omg.org/spec/DMN/20180521/DC/\" id=\"Definitions_1yu2lcr\" name=\"DRD\" namespace=\"http://camunda.org/schema/1.0/dmn\" exporter=\"Camunda Modeler\" exporterVersion=\"4.11.1\">\n" +
            "  <decision id=\"item\" name=\"item not resolved demo\">\n" +
            "    <variable id=\"InformationItem_05x2m7o\" name=\"result\" typeRef=\"string\" />\n" +
            "    <literalExpression id=\"LiteralExpression_0g7w721\" expressionLanguage=\"feel\">\n" +
            "      <text>[1,2,3,4,5][custom_even(item)]</text>\n" +
            "    </literalExpression>\n" +
            "  </decision>\n" +
            "</definitions>";

    public void run() {
        DefaultDmnEngineConfiguration dmnEngineConfig = (DefaultDmnEngineConfiguration) DmnEngineConfiguration.createDefaultDmnEngineConfiguration();
        dmnEngineConfig.feelCustomFunctionProviders(Collections.singletonList(new CustomFuncProvider()));

        DmnEngine dmnEngine = dmnEngineConfig.buildEngine();

        DmnModelInstance dmnModel = Dmn.readModelFromStream(new ByteArrayInputStream(DMN.getBytes(StandardCharsets.UTF_8)));

        DmnDecision item = dmnEngine.parseDecision("item", dmnModel);
        dmnEngine.evaluateDecision(item, new HashMap<>());
    }

    public static class CustomFuncProvider implements FeelCustomFunctionProvider {
        private final Map<String, CustomFunction> functions = new HashMap<>();

        public CustomFuncProvider() {
            functions.put("custom_even", CustomFunction.create()
                    .setParams("val")
                    .setFunction(args -> {
                        System.out.println(args.get(0) + " class=" + args.get(0).getClass().getName());
                        return true;
                    }).build());
        }

        @Override
        public Optional<CustomFunction> resolveFunction(String key) {
            return Optional.ofNullable(functions.get(key));
        }

        @Override
        public Collection<String> getFunctionNames() {
            return functions.keySet();
        }
    }
}
```
Executing the run method produces the following output:
```
ValError(no variable found for name 'item') class=org.camunda.feel.syntaxtree.ValError
1 class=java.lang.Long
2 class=java.lang.Long
3 class=java.lang.Long
4 class=java.lang.Long
5 class=java.lang.Long
```

**Expected behavior**
The custom function should be invoked for each list element and should not be invoked with ValError.
On the other hand there is some ambiguity here - item variable can be defined or not defined in the outside scope, the function return value could be boolean or integer. Maybe it is safe to assume that if "item" variable is used inside [ ] then the expression in [ ] should be invoked for each list element and return a boolean or fail otherwise, except for cases like this:
```
  [10,20,30,40][    ([1,2][even(item)])[1]   ]  // expected 20
``` 

**Environment**
Standalone Java App with Camunda DMN engine:
```xml
        <dependency>
            <groupId>org.camunda.bpm.dmn</groupId>
            <artifactId>camunda-engine-dmn-bom</artifactId>
            <version>7.16.0</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
```

saig0:
## Description

* a list can be filtered using a boolean function
* currently, the function is invoked once to determine if it is a number (for index access) or a boolean value (or something else)
* since the function is invoked without the list item, the invocation can fail if the "item" variable is passed in as a parameter
* fix the issue by avoiding the function invocation without "item" variable in the context

## Related issues

closes #359

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -253,21 +253,23 @@ class FeelInterpreter {
       case Filter(list, filter) =>
         withList(
           eval(list),
-          l =>
+          l => {
+            val evalFilterWithItem =
+              (item: Val) => eval(filter)(filterContext(item))
+
             filter match {
               case ConstNumber(index) => filterList(l.items, index)
               case ArithmeticNegation(ConstNumber(index)) =>
                 filterList(l.items, -index)
-              case comparison: Comparison =>
-                filterList(l.items,
-                           item => eval(comparison)(filterContext(item)))
+              case _: Comparison | _: FunctionInvocation |
+                  _: QualifiedFunctionInvocation =>
+                filterList(l.items, evalFilterWithItem)
               case _ =>
                 eval(filter) match {
                   case ValNumber(index) => filterList(l.items, index)
-                  case _ =>
-                    filterList(l.items,
-                               item => eval(filter)(filterContext(item)))
+                  case _                => filterList(l.items, evalFilterWithItem)
                 }
+            }
           }
         )
       case IterationContext(start, end) =>
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterListExpressionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterListExpressionTest.scala
@@ -21,6 +21,8 @@ import org.camunda.feel.syntaxtree._
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.flatspec.AnyFlatSpec
 
+import scala.collection.mutable.ListBuffer
+
 /**
   * @author Philipp Ossler
   */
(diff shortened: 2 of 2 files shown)


pihme (approved):
Looks good to me :tada:

PR #423, opened 2022-05-06
Expose public getters of a POJO

### LINKED ISSUE #360: Public getters of POJOs are not exposed to custom functions
**Describe the bug**
When Java object is passed to a decision and then from the decision to a custom FEEL function, in the Java body of the function we get empty instance of camundajar.impl.scala.collection.convert.JavaCollectionWrappers$MapWrapper, not the original Java object (or even some kind of proxy map).

**To Reproduce**
The following code snipped can be used to reproduce the bug:
```java
public class DmnEngineTest2 {
    static final String DMN = "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
            "<definitions xmlns=\"https://www.omg.org/spec/DMN/20191111/MODEL/\" xmlns:dmndi=\"https://www.omg.org/spec/DMN/20191111/DMNDI/\" xmlns:dc=\"http://www.omg.org/spec/DMN/20180521/DC/\" id=\"Definitions_1yu2lcr\" name=\"DRD\" namespace=\"http://camunda.org/schema/1.0/dmn\" exporter=\"Camunda Modeler\" exporterVersion=\"4.11.1\">\n" +
            "  <decision id=\"pojo\" name=\"custom function with Java pojo (literal)\">\n" +
            "    <variable id=\"InformationItem_0zmcpc8\" name=\"result\" typeRef=\"string\" />\n" +
            "    <literalExpression id=\"LiteralExpression_1gccogk\" expressionLanguage=\"feel\">\n" +
            "      <text>custom_pojo(pojo_input)</text>\n" +
            "    </literalExpression>\n" +
            "  </decision>\n" +
            "</definitions>";

    public void run() {
        DefaultDmnEngineConfiguration dmnEngineConfig = (DefaultDmnEngineConfiguration) DmnEngineConfiguration.createDefaultDmnEngineConfiguration();
        dmnEngineConfig.feelCustomFunctionProviders(Collections.singletonList(new CustomFuncProvider()));

        DmnEngine dmnEngine = dmnEngineConfig.buildEngine();

        DmnModelInstance dmnModel = Dmn.readModelFromStream(new ByteArrayInputStream(DMN.getBytes(StandardCharsets.UTF_8)));

        DmnDecision decision = dmnEngine.parseDecision("pojo", dmnModel);

        Map<String, Object> variables = new HashMap<>();
        variables.put("pojo_input", new Pojo());
        dmnEngine.evaluateDecision(decision, variables);
    }

    public static class CustomFuncProvider implements FeelCustomFunctionProvider {
        private final Map<String, CustomFunction> functions = new HashMap<>();

        public CustomFuncProvider() {
            functions.put("custom_pojo", CustomFunction.create()
                    .setParams("val")
                    .setFunction(args -> {
                        System.out.println(args.get(0) + " class=" + args.get(0).getClass().getName());
                        Map map = (Map) args.get(0);
                        System.out.println("list=" + map.get("list") + " num=" + map.get("num"));
                        return map.get("num");
                    }).build());
        }

        @Override
        public Optional<CustomFunction> resolveFunction(String key) {
            return Optional.ofNullable(functions.get(key));
        }

        @Override
        public Collection<String> getFunctionNames() {
            return functions.keySet();
        }
    }

    public static class Pojo {
        private int num = 10;
        private List<Integer> list = Arrays.asList(1, 2, 3);

        public int getNum() {
            return num;
        }

        public List<Integer> getList() {
            return list;
        }
    }
}
```
When the run method is executed the following output is printed:
```
{} class=camundajar.impl.scala.collection.convert.JavaCollectionWrappers$MapWrapper
list=null num=null
```

**Expected behavior**
~In the body of the custom_pojo function I would expect to receive the original Java object or at least some kind of Map-proxy with non-empty keys and values.~

The context that is passed to the custom function contains entries for all public fields and for all **public getters**.

**Environment**
Standalone Java app with the Camunda DMN engine:
```
        <dependency>
            <groupId>org.camunda.bpm.dmn</groupId>
            <artifactId>camunda-engine-dmn-bom</artifactId>
            <version>7.16.0</version>
            <type>pom</type>
            <scope>import</scope>
        </dependency>
```

saig0:
## Description

* expose public getters of a POJO
* the getter must be public and match to a field (i.e.`getNameOfField()` or `isNameOfField()`)

## Related issues

closes #360

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/ObjectContext.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/ObjectContext.scala
@@ -19,6 +19,8 @@ package org.camunda.feel.impl.interpreter
 import org.camunda.feel.context.{Context, FunctionProvider, VariableProvider}
 import org.camunda.feel.syntaxtree.ValFunction
 
+import java.lang.reflect.Method
+
 /**
   * A context that wraps the fields and methods of a given JVM object
   *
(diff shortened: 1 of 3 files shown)


remcowesterhoud (approved):
Nice! Good to see the refactoring at the end because I had a bunch of comments ready for this 😄 

LGTM

PR #426, opened 2022-05-12
docs: add caution to list index

korthout:
## Description

This cautionary note is already available on the Get Element section of List Expressions.
However, as users may run into this early, it might make sense to highlight this oddity
also on the data types page.

### DIFF
--- a/docs/docs/reference/language-guide/feel-data-types.md
+++ b/docs/docs/reference/language-guide/feel-data-types.md
@@ -153,6 +153,10 @@ A list of elements. The elements can be of any type. The list can be empty.
 [["list"], "of", [["lists"]]]
 ```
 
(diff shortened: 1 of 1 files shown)


saig0:
First of all, I'm very happy about your interest in improving the docs! 

For this specific change, I don't think that it fits because this note is already on the list expressions page. I don't expect the user here to look for this information. I'm sorry :bow: 

However, I love your idea about the new learning guide :+1:

saig0:
First of all, I'm very happy about your interest in improving the docs! 

For this specific change, I don't think that it fits because this note is already on the list expressions page. I don't expect the user here to look for this information. I'm sorry :bow: 

However, I love your idea about the new learning guide :+1:

PR #434, opened 2022-06-15
Return time with zone id/offset as OffsetTime

### LINKED ISSUE #163: Return a time value with zone-id as OffsetTime
When time value with a zone-id (e.g. `time("10:31:10@Europe/Paris")`) is returned then it has the type `ZonedTime`. This type is a special type of the FEEL engine and not part of the normal Java/Scala classes. This introduces a dependency to the engine code, forces the developers to look into the code, and may cause problems when integrating it into the application code. 

A time value with zone-id is returned as `java.time.OffsetTime` instead. 
The zone-id component is ignored. The offset is created of the standard offset of the given zone-id at the current point in time.

**Related issues**

saig0:
## Description

* return a time with zone id/offset as `java.time.OffsetTime` instead of custom `ZonedTime`

## Related issues

closes #163


remcowesterhoud (approved):
👍 LGTM

PR #435, opened 2022-06-15
Parse unary-tests expression with boolean and conjunction/disjunction 

### LINKED ISSUE #429: Boolean Expressions: Parsing error when using Conjunction ("And") or Disjunction ("Or")
**Describe the bug**

We use FEEL in Camunda's DMN tables to evaluate expressions.
After we upgraded from Camunda `7.16.7` (with FEEL `1.13.3`) to `7.17.1` (FEEL `1.14.2`) we discovered a parsing error in FEEL when it comes to evaluating boolean expressions.

image

`FEEL/SCALA-01008 Error while evaluating expression: failed to parse expression 'true and null': Expected ("," | end-of-input):1:6, found "and null"`

Before the upgrade the expression "true and null" could be evaulated without problems and resulted in "null" (which is in line with https://camunda.github.io/feel-scala/docs/1.14/reference/language-guide/feel-boolean-expressions/#conjunction--and).

When the expression is enclosed in paretheses tha parsing error does not appear. So this is OK: `(true and null)`
The same problem also arises when using `or` in a boolean expression, e.g. `true or true`

**To Reproduce**
1. Create a DMN table in Camunda modeler like in the screenshot above.
2. Call `DmnDecisionTableResult.getDecisionResult(..)`

**Expected behavior**
No parsing error should occur when two booleans are combined with `and` or `or`.

**Environment**
* FEEL engine version: [1.14.2]
* Affects: 
  * Camunda Automation Platform 7: [7.17.1]

saig0:
## Description

* a unary-tests expression with a boolean literal and a conjunction/disjunction failed to parse
* the parser expected no other chars (EOF) after the boolean literal was accepted
* adjust the parser to accept the boolean literal only at the end of the expression (EOF)

## Related issues

closes #429

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -574,7 +574,7 @@ object FeelParser {
 
   // boolean literals are ambiguous for unary-tests. give precedence to comparison with input.
   private def positiveUnaryTest[_: P]: P[Exp] =
-    boolean.map(InputEqualTo) | expression.map(UnaryTestExpression)
+    (boolean.map(InputEqualTo) ~ End) | expression.map(UnaryTestExpression)
 
   private def anyInput[_: P]: P[Exp] =
     P(
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
@@ -185,12 +185,27 @@ class InterpreterUnaryTest
   it should "compare to a conjunction (and)" in {
     evalUnaryTests(true, "true and true") shouldBe ValBoolean(true)
     evalUnaryTests(true, "false and true") shouldBe ValBoolean(false)
+
+    evalUnaryTests(true, "true and null") shouldBe ValBoolean(false)
+    evalUnaryTests(true, "false and null") shouldBe ValBoolean(false)
+
+    evalUnaryTests(true, """true and "otherwise" """) shouldBe ValBoolean(
+      false)
+    evalUnaryTests(true, """false and "otherwise" """) shouldBe ValBoolean(
+      false)
   }
 
   it should "compare to a disjunction (or)" in {
     evalUnaryTests(true, "true or true") shouldBe ValBoolean(true)
     evalUnaryTests(true, "false or true") shouldBe ValBoolean(true)
     evalUnaryTests(true, "false or false") shouldBe ValBoolean(false)
+
+    evalUnaryTests(true, "true or null") shouldBe ValBoolean(true)
+    evalUnaryTests(true, "false or null") shouldBe ValBoolean(false)
+
+    evalUnaryTests(true, """true or "otherwise" """) shouldBe ValBoolean(true)
+    evalUnaryTests(true, """false or "otherwise" """) shouldBe ValBoolean(
+      false)
   }
 
   "A date" should "compare with '<'" in {
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
@@ -182,6 +182,17 @@ class InterpreterUnaryTest
     evalUnaryTests(true, """ "a" = "b" """) should be(ValBoolean(false))
   }
 
+  it should "compare to a conjunction (and)" in {
+    evalUnaryTests(true, "true and true") shouldBe ValBoolean(true)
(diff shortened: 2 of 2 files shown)


korthout (changes requested):
I'm not sure this correctly solves the issue, although that might just be my misunderstanding of the parser. 

❌ Nevertheless, there is no test that shows that the regression is resolved.

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
```diff
@@ -182,6 +182,17 @@ class InterpreterUnaryTest
...
+    evalUnaryTests(true, "true and true") shouldBe ValBoolean(true)
+    evalUnaryTests(true, "false and true") shouldBe ValBoolean(false)
+  }
+
+  it should "compare to a disjunction (or)" in {
+    evalUnaryTests(true, "true or true") shouldBe ValBoolean(true)
+    evalUnaryTests(true, "false or true") shouldBe ValBoolean(true)
+    evalUnaryTests(true, "false or false") shouldBe ValBoolean(false)
+  }
+
```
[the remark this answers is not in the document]
saig0: Fair point :+1: 

> IMO, they should be tested as it is documented behavior.

There are tests :sweat_smile: But for expressions. It is unusual that this is used for unary-tests in this way.

However, I added additional test cases :heavy_check_mark:

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
```diff
@@ -574,7 +574,7 @@ object FeelParser {
 
   // boolean literals are ambiguous for unary-tests. give precedence to comparison with input.
   private def positiveUnaryTest[_: P]: P[Exp] =
-    boolean.map(InputEqualTo) | expression.map(UnaryTestExpression)
+    (boolean.map(InputEqualTo) ~ End) | expression.map(UnaryTestExpression)
```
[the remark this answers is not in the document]
saig0: > Doesn't true now require 2 EOF markers?

saig0:
@korthout please have another look. I added the missing test cases.

src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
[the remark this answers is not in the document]
korthout: Thanks for the explanation. I understood the backtracking but imagined consumption of the tokens somehow. It makes much more sense to me now. 👍

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
```diff
@@ -185,12 +185,27 @@ class InterpreterUnaryTest
   it should "compare to a conjunction (and)" in {
     evalUnaryTests(true, "true and true") shouldBe ValBoolean(true)
     evalUnaryTests(true, "false and true") shouldBe ValBoolean(false)
+
+    evalUnaryTests(true, "true and null") shouldBe ValBoolean(false)
```
korthout, line 189: ❌ This case is wrong, the result should be `null` (there are more such incorrect cases added in this commit)

```
true and null
// null
```

See [docs](https://camunda.github.io/feel-scala/docs/1.14/reference/language-guide/feel-boolean-expressions/#conjunction--and)

korthout (changes requested):
Thanks for adding the test cases

❌ Some of the cases don't reflect the documented behavior

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
saig0: Sorry, but the cases are correct :sweat_smile: In the docs, it is about expressions but the test cases are about unary-tests :mag: 

The difference is that a unary-tests is compared to the input value.  So, `true and null` evaluates to `null`, and then it compares to `null` with `true` that result in `false`.

saig0: Happy to chat about it :speaking_head:

korthout: >So, true and null evaluates to null, and then it compares to null with true that result in false.

I see now how it works: the last comparison is due to it being a unary test. It compares the input (`true`) with the result of the expression `null`. These are not the same, so the result is `false`.

You could argue that this test does not verify the behavior of evaluating the unary test, e.g. `true and null` is also not equal to "random string". But you mentioned that verifying the evaluation was not the intent of this test case. The intent is to verify that it can be parsed (this was the bug). Users should generally not use this `true and null` expression as a unary test.

korthout (approved):

PR #440, opened 2022-06-30
Dependabot analyse GHA workflows

remcowesterhoud:
## Description

* Bump add-to-project to version v0.1.0
* Have dependabot create PR's for outdated actions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,22 @@
 version: 2
 updates:
-- package-ecosystem: maven
-  directory: "/"
-  schedule:
-    interval: daily
-    timezone: Europe/Berlin
-  ignore:
-  - dependency-name: org.antipathy:mvn-scalafmt_2.12
-    versions:
-    - ">= 0.9_1.a, < 0.9_1.1000004"
-  - dependency-name: org.antipathy:mvn-scalafmt_2.12
-    versions:
-    - ">= 1.0.a, < 1.1"
-  - dependency-name: com.mycila:license-maven-plugin
-    versions:
-    - "4.0"
+  - package-ecosystem: maven
+    directory: "/"
+    schedule:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -1,17 +1,22 @@
...
+        versions:
+          - ">= 1.0.a, < 1.1"
+      - dependency-name: com.mycila:license-maven-plugin
+        versions:
+          - "4.0"
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
korthout, line 22: ❓ For the `maven` ecosystem the schedule is set to a specific timezone. I'm not sure what the benefit is, but should we do the same for the GHA ecosystems?

.github/dependabot.yml
```diff
@@ -1,17 +1,22 @@
...
-    versions:
-    - ">= 0.9_1.a, < 0.9_1.1000004"
-  - dependency-name: org.antipathy:mvn-scalafmt_2.12
-    versions:
-    - ">= 1.0.a, < 1.1"
-  - dependency-name: com.mycila:license-maven-plugin
-    versions:
-    - "4.0"
+  - package-ecosystem: maven
+    directory: "/"
```
korthout, line 4: ❓ Any particular reason why the indentation changed? Was it not correctly formatted yaml? Is it currently formatted correctly?

remcowesterhoud: Exactly, it wasn't properly formatted before.

.github/dependabot.yml
remcowesterhoud: It ha to do with the time dependabot start analysing. I'm not sure what the default start time is, but it really doesn't matter. It runs once a day which is fine. I don't care what time it starts.

PR #485, opened 2022-08-10
Update Docusaurus to 2.0

korthout:
## Description

Apart from updating to the newer stable docusaurus 2.0, this also fixes the DocSearch indexing via Angolia.

* updates `@docusaurus/core` and `@docusaurus/presets-classic` to the latest version: `2.0.1`
* updates the package-lock file to v2 format
* configures the new Angolia app settings, which re-enables indexing of the docs for DocSearch (it seems the index was frozen and no longer updated)
* updates GHA workflows to use node v18, to fix the doc CI

## Related issues

NA


saig0 (approved):
@korthout we're up-to-date again. Thank you :rocket:

PR #498, opened 2022-08-12
Handle context errors elegantly

korthout:
## Description

* If the context is not a correct JSON object, then we display an error message.
* If possible it also shows the position where the error
occurred.

## Related issues

Summer hackdays


saig0 (approved):
Nice :+1:

PR #505, opened 2022-08-21
docs: fix #504, playground can't display legitimate false return values

### LINKED ISSUE #504: Playground can't display legitimate `false` return values
## Which documentation is missing/incorrect?

The playground should display `false` if an expression evaluates to `false`.  The current implementation assumes a `result` of `false` indicates an error. However, the result can also legitimately be `false` (without an error). 

Link to the documentation:
* https://camunda.github.io/feel-scala/docs/playground/

Affected versions:
* 1.16.0-SNAPSHOT

upgradingdave:
## Description

* Small change to `LiveFeel.js` to look at "error" rather than "result" in order to know whether or not to display error. This way, expressions that legitimately evaluate to `false` will be displayed in playground results. 

## Related issues

closes #504

### DIFF
--- a/docs/src/components/LiveFeel.js
+++ b/docs/src/components/LiveFeel.js
@@ -67,9 +67,7 @@ const LiveFeel = ({
         }
       )
       .then((response) => {
-        if (response?.data?.result) {
-          onResult(JSON.stringify(response.data.result));
-        } else if (response?.data?.error) {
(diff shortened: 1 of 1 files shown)


saig0 (approved):
@upgradingdave looks good :+1: 

Thank you for fixing this issue :cookie:

PR #538, opened 2022-10-13
(feat): Fix negative durations in string variables that fail to parse

### LINKED ISSUE #513: negative durations in string variables fail to parse
negative durations work when they are literal strings
`feel("""duration("PT-5M")""", context)` works

however when they are string variables they don't 

`feel("""duration(durationBefore)""", Map("durationBefore" -> "PT-5M"))` shows error

```
12:42:40.918 [main] WARN  org.camunda.feel.FeelEngine - Failed to parse duration from 'PT-5M'
```

sccalabr:
## Description

* 

Negative durations when they are string variables don't parse

feel("""duration(durationBefore)""", Map("durationBefore" -> "PT-5M")) shows error

This pr fixes the issue. 
## Related issues

closes #513

### DIFF
--- a/src/main/scala/org/camunda/package.scala
+++ b/src/main/scala/org/camunda/package.scala
@@ -120,7 +120,7 @@ package object feel {
     localDateTimePatter.matcher(dateTime).matches
 
   private lazy val yearMonthDurationPattern =
-    Pattern.compile("""-?P(\d+Y)?(\d+M)?""")
+    Pattern.compile("""-?P(\d+Y)?-?(\d+M)?""")
 
   def isYearMonthDuration(duration: String): Boolean =
     yearMonthDurationPattern.matcher(duration).matches
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
@@ -207,6 +207,12 @@ class BuiltinConversionFunctionsTest
       ValDayTimeDuration("P2DT20H14M"))
   }
 
+  "A duration() function" should "parse string variables with -" in {
+
+    eval(""" duration(x) """, Map("x" -> "PT-5M")) should be(
+      ValDayTimeDuration("PT-5M"))
+  }
+
   it should "convert year-month-String" in {
 
     eval(""" duration(x) """, Map("x" -> "P2Y4M")) should be(


saig0:
@sccalabr thank you for your contribution. :tada:  I'll have a look at your PR in the next few days. :eyes:

sccalabr:
Does this look good?

src/main/scala/org/camunda/package.scala
```diff
@@ -120,13 +120,13 @@ package object feel {
...
   private lazy val yearMonthDurationPattern =
-    Pattern.compile("""-?P(\d+Y)?(\d+M)?""")
+    Pattern.compile("""-?P(\d+Y)?-?(\d+M)?""")
 
   def isYearMonthDuration(duration: String): Boolean =
     yearMonthDurationPattern.matcher(duration).matches
 
   private lazy val dayTimeDurationPattern =
-    Pattern.compile("""-?P((\d+D)?(T(\d+H)?(\d+M)?(\d+(.\d*)?S)?)?)""")
+    Pattern.compile("""-?P((\d+D)?(T(\d+H)?-?(\d+M)?(\d+(.\d*)?S)?)?)""")
```
saig0, line 129: :x: We need to adjust the pattern a bit more.

The minus sign could be placed on each component. Currently, it doesn't work for:
- `P-1D`
- `PT-2H`
- `PT-3M-4S`

src/main/scala/org/camunda/package.scala
```diff
@@ -120,13 +120,13 @@ package object feel {
     localDateTimePatter.matcher(dateTime).matches
 
   private lazy val yearMonthDurationPattern =
-    Pattern.compile("""-?P(\d+Y)?(\d+M)?""")
+    Pattern.compile("""-?P(\d+Y)?-?(\d+M)?""")
```
saig0, line 123: :x: We need to adjust the pattern a bit more.

The minus sign could be placed on each component. Currently, it doesn't work for:
- `P-1Y`

src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
```diff
@@ -207,6 +207,12 @@ class BuiltinConversionFunctionsTest
       ValDayTimeDuration("P2DT20H14M"))
   }
 
+  "A duration() function" should "parse string variables with -" in {
+
+    eval(""" duration(x) """, Map("x" -> "PT-5M")) should be(
+      ValDayTimeDuration("PT-5M"))
```
saig0, line 213: :x: We should add more test cases to cover all possible options.

For example:

- `P-1DT-2H-3M-4S` - returns `null`
- `-P1DT2H3M4S` - works

saig0: :wrench: We could add similar test cases for a years-months-duration string. See the test case below.

saig0: :wrench: We could add similar test cases for the duration literals (i.e. `duration("-P1D")` vs `duration(x)`).

The test cases for the literals are in the class [InterpreterLiteralExpressionTest](https://github.com/camunda/feel-scala/blob/main/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterLiteralExpressionTest.scala#L236-L262). Currently, they don't cover negative durations.

saig0 (changes requested):
@sccalabr your changes fix the reported issue. :+1: 

However, I found more cases that are not covered yet. Please have a look at my comments. :cookie:

src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
sccalabr: Should the InterpreterTest be a follow up work item? Seems like that would be a different issue right?

saig0: > Should the InterpreterTest be a follow up work item? Seems like that would be a different issue right?

Yes. This was not part of the reported issue. But the implementation uses the same code. 

Feel to adjust the `InterpreterLiteralExpressionTest` in a follow-up PR.

saig0 (approved):
@sccalabr your PR is ready to merge. :rocket: 

Thank you for applying the hints. :cookie: 
Feel free to adjust the `InterpreterLiteralExpressionTest` in a follow-up PR. 

---

This pull request counts toward the Hacktoberfest challenge. If you contributed four pull requests then you can complete your Camunda Hacktoberfest challenge [here](https://camunda.com/hacktoberfest2022/).

Issue #308, opened 2021-08-03
UnaryTests: compare a list with a single value


saig0:
**Is your feature request related to a problem? Please describe.**
In a unary-tests, I want to compare a variable with the input value. If the variable evaluates to a list then it fails to evaluate the expression because it expects a single value.

It works if the variable is replaced with a literal of the same value.

![image](https://user-images.githubusercontent.com/4305769/127957972-bd98e2ff-a1e0-46a6-aeee-c741362f0656.png)

**Describe the solution you'd like**
In a unary-tests, I can compare a variable with the input value if the variable evaluates to a list.

**Related issues**
Related to #307.

* Camunda BPM: 
* Zeebe broker:

sccalabr:
Id also like to give this a try

saig0:
The issue is fixed since version `1.14.0`.

PR #548, opened 2022-11-03
test: verify unary-tests with list variable

saig0:
## Description

Add a test case to verify that a value in a unary-tests expression can be compared to a variable with a list value.

## Related issues

Verifies that #308 is fixed (since version `1.14.0`).


korthout (approved):
Nice and concise 💯 

LGTM 👍 

bors merge

korthout:
Ah, no bors here

PR #550, opened 2022-11-03
docs: Mark Camunda Extensions

### LINKED ISSUE #524: Mark features that are not part of the current DMN spec
## Which documentation is missing/incorrect?

We should mark all features or built-in functions that are currently not part of the specification. 

### List of features

* Language Features
    * [ ] <code>\`backtick variable escapes\`</code> https://camunda.github.io/feel-scala/docs/reference/language-guide/feel-variables/#escape-variable-names
* Custom Functions
    * [ ] `is defined` https://camunda.github.io/feel-scala/docs/reference/builtin-functions/feel-built-in-functions-boolean/#is-defined
    * [ ] ...
    
This requirement was brought up [here](https://camunda.slack.com/archives/C01QYD808A3/p1664185992704599).

saig0:
## Description

IN PROGRESS. :construction: 

## Related issues

closes #524


saig0:
@korthout please have a look at my PR. :cookie:

nikku:
Looks good to me :heart:.

I was in the process to add more detail to https://github.com/camunda/feel-scala/issues/524 when I saw that you already fixed places relevant to me. :medal_sports:

docs/src/components/MarkerCamundaExtension.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export const MarkerCamundaExtension = () => {
+  return (
+      <p>
+        <span style={{
+          backgroundColor: '#25c2a0',
```
korthout, line 7: 🔧 You could align this with the brand color codes, e.g. Greenmunda: `#26D07C`

docs/src/components/MarkerChangelogVersion.js
```diff
@@ -0,0 +1,26 @@
...
+  return (
+      <p>
+        <span style={{
+          backgroundColor: '#25c2a0',
+          borderRadius: '7px',
+          color: '#fff',
+          padding: '0.2rem',
+          marginRight: '0.5rem'
+        }}
+              title={"Available since the given Camunda Platform 8 (Zeebe) version."}>Zeebe: {versionZeebe}</span>
```
korthout, line 14: 🔧 Might be better to rename `Zeebe: ` to `Camunda Platform 8: `

docs/src/components/MarkerChangelogVersion.js
```diff
@@ -0,0 +1,26 @@
...
+        <span style={{
+          backgroundColor: '#1877F2',
```
korthout, line 16: 🔧 Same point about aligning, but my suggestion is `#0072CE`

korthout (approved):
Awesome idea to use MDX components for this!

LGTM 👍 I just have a few suggestions

docs/src/components/MarkerCamundaExtension.js
nikku: If we discuss colors / align this I'd rather use a neutral or warning color than a "awesome, go for this" color, cf. [here](https://miro.com/app/board/uXjVPOXAyQ4=/?moveToWidget=3458764536100903050&cot=14). We don't want to extra "encourage" use of vendor extensions. Some may be safe to adopt across our tooling (i.e. adding a custom function), others will make you run into trouble (i.e. the not FEEL spec compliant backtick string).

saig0: I'll change the color to `#FC5D0D` (Orangemunda).

PR #563, opened 2022-11-21
docs: Update version banner

saig0:
## Description

Set the last version to the last released version, instead of the current dev version. This changes the behavior to redirect by default to the last released version. In most cases, this should be the version that users are looking for.

Disable the "unmaintained version" for the last released version (1.15) and all supported versions (1.14).

For reference: https://docusaurus.io/docs/versioning#configuring-versioning-behavior.

---

## Before

image

--- 

## After

image

image

image

image

---


remcowesterhoud (approved):
Nice improvements 👍 

bors merge

remcowesterhoud:
Of course, no Bors 😄

PR #601, opened 2023-03-10
Add new get value function with dynamic path

koevskinikola:
## Description

Add a new get value function with a new signature. With this function, a user can dynamically pass a path to obtain a nested property in a given context.

## Related issues

closes [#11293](https://github.com/camunda/zeebe/issues/11293)

### DIFF
--- a/docs/docs/reference/builtin-functions/feel-built-in-functions-context.md
+++ b/docs/docs/reference/builtin-functions/feel-built-in-functions-context.md
@@ -26,6 +26,33 @@ get value({a: 1}, "b")
 // null
 ```
 
+## get value(context, keys)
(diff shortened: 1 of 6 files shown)


saig0 (changes requested):
@koevskinikola thank you for driving our mob programming session. :rocket: 

:x: I found one case that could lead to a cast exception in Zeebe. 
:x: We need to document the new function. 
:wrench: I have an idea to refactor the function a bit.

Please have a look at my comments. Happy to answer your questions or discuss/code f2f. :cookie:

saig0 (approved):
@koevskinikola LGTM. :tada:

PR #639, opened 2023-05-12
build: Ignore inner methods for Clirr

saig0:
## Description

Clirr reports a false positive if an inner method is changed. Adjust the configuration of the Clirr plugin to ignore these errors.


koevskinikola (approved):
LGTM :+1:

PR #653, opened 2023-06-01
fix: Preserve the order of a context

### LINKED ISSUE #161: get entries() function returns the entries in a different order

saig0:
## Description

Fix an issue that did not store the context entries in the same order as defined in the literal (i.e. `{a: 1, b: 2}` is stored as `{b: 2, a: 1}`). The issue could be observed by using the `get entries()` function.

The root cause was not in the context functions but in how the context is created in the engine via `EvalContext`. Fixing the behavior by using a sorted map for storing the context entries and by avoiding the nesting of the `VariableProvider`s.

As a result, the code of the `EvalContext` is a bit more complex but easier to debug. All variables and functions are stored in one provider instead of a chain with one provider per variable/function.

## Related issues

closes #161 
based on a previous try #340

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
@@ -24,7 +24,7 @@ import org.camunda.feel.impl.interpreter.EvalContext.{mergeFunctionProviders, me
 import org.camunda.feel.syntaxtree.{Val, ValError, ValFunction}
 import org.camunda.feel.valuemapper.ValueMapper
 
-import scala.collection.immutable.SortedMap
+import scala.collection.immutable.SeqMap
 
 class EvalContext(val valueMapper: ValueMapper,
                   val variableProvider: VariableProvider,
@@ -162,9 +162,9 @@ object EvalContext {
   }
 
   private def toSortedVariableProvider(variables: Map[String, Any]): VariableProvider =
-    StaticVariableProvider(SortedMap[String, Any]() ++ variables)
+    StaticVariableProvider(SeqMap[String, Any]() ++ variables)
 
   private def toSortedVariableProvider(entry: (String, Any)): VariableProvider =
-    StaticVariableProvider(SortedMap(entry))
+    StaticVariableProvider(SeqMap(entry))
 
 }
--- a/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
@@ -16,16 +16,19 @@
  */
 package org.camunda.feel.impl.interpreter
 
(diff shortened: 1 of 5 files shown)


src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
```diff
@@ -35,77 +38,133 @@ class EvalContext(val valueMapper: ValueMapper,
...
+        val functionsByKey = (key: String) => thisFunctions.getOrElse(key, List.empty) ++ otherFunctions.getOrElse(key, List.empty)
+        val allFunctions = allKeys.map(key => key -> functionsByKey(key)).toMap
+        StaticFunctionProvider(allFunctions)
+      }
+      case (thisProvider, otherProvider) => CompositeFunctionProvider(List(thisProvider, otherProvider))
+    }
+  }
+
+  private def toSortedVariableProvider(variables: Map[String, Any]): VariableProvider =
+    StaticVariableProvider(SortedMap[String, Any]() ++ variables)
```
remcowesterhoud, line 165: ❌ As discussed f2f, this [SortedMap](https://www.scala-lang.org/api/2.13.6/scala/collection/immutable/SortedMap.html) does not behave the way you think it does. It doesn't keep the entries sorted in the order they were added. Instead they are sorted by their key, regardless of when they're added to the map.

Let's also add a test case for this scenario 🙂

src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
```diff
@@ -35,77 +38,133 @@ class EvalContext(val valueMapper: ValueMapper,
...
+
+  private def toSortedVariableProvider(entry: (String, Any)): VariableProvider =
+    StaticVariableProvider(SortedMap(entry))
```
remcowesterhoud, line 168: ❌ Same as above

remcowesterhoud (changes requested):
Thanks for hopping on a call to explain the changes further @saig0! Please have a look at my comments!

saig0:
@remcowesterhoud great catch. :+1: 

I fixed the issue by using a [SeqMap](https://www.scala-lang.org/api/2.13.6/scala/collection/SeqMap.html). :laughing: Additionally, I added a few more test cases with non-alphabetic order. :see_no_evil: 

Please have a look. :eyes:

remcowesterhoud (approved):
Thanks @saig0! 🚀

remcowesterhoud:
bors merge

PR #673, opened 2023-07-06
[WIP]: Customize the error handling

### LINKED ISSUE #260: Customize the error handling
FEEL doesn't have strict error handling. An expression may evaluate to `null` (and continue). For example, if the types don't match or a function is called with the wrong arguments.  

For example:
```
"order-" + 123
// WARN  org.camunda.feel.FeelEngine - Suppressed failure: expected String but found 'ValNumber(123)'
// should be "order-" + string(123)

date("01-01-2021")
// null - wrong format
// should be date("2021-01-01")
```

In some cases, you want to have different behavior. For example, collect and report all (suppressed) failures. This may help to identify failures and to correct the expression. 

Have the option to customize the error handling. ~For example, using an SPI like for the value mapper and the function provider.~ 

A solution could be to give control of the error handling to the caller of the API. 

If the engine can't evaluate a part of the expression successfully then it collects the failure and continues the evaluation with `null`. It may return `null` as the result of the evaluation and exposes the collected failures (i.e. failure message + failure type). 

Based on the collected failures, the caller can decide if the evaluation was successful or not.  

**Related issues**

Kind of related to #259

saig0:
## Description

:construction: In progress :construction: 

- [x] Extract and extend failure types
- [x] Moprh ValError into suppressed failures
- [x] Create a new API for evaluation to return the full result
- [ ] Create a follow-up issue for strict engine mode 
- [ ] Create a follow-up issue for a new debug built-in function (`debug(3 + x).failures`) 

## Related issues

closes #260


saig0:
Close in favor of https://github.com/camunda/feel-scala/pull/687.

Issue #582, opened 2022-12-22
Handle non-existing variables and context entries gracefully in comparison and filter expressions


saig0:
**Describe the bug**
If I compare a non-existing variable with a value then the expression fails. To avoid failure, I must do a null check first.

```
x = 3
// failed to evaluate expression ' x = 3 ': no variable found for name 'x'

x > 3
// failed to evaluate expression ' x > 3 ': ValError(no variable found for name 'x') is not comparable

x != null and x = 3
// returns false
```

The behavior is the same for filter expressions.

```
[{x:5},{y:2}][x > 3]
// failed to evaluate expression ' [{x:5},{y:2}][x > 3]  ': ValError(no variable found for name 'x') is not comparable)

[{x:5},{y:2}][x != null and x > 3]
// returns [{x:5}]
```

Related [discussion](https://camunda.slack.com/archives/C01QYD808A3/p1671687820836529).
Related to https://github.com/camunda/feel-scala/issues/572.
Related to https://github.com/camunda/feel-scala/issues/260.

**To Reproduce**
1. Evaluate the expression `x = 5`
2. Verify that the expression is not successful

**Expected behavior**
A non-existing variable or context entry is handled gracefully in a comparison and a filter.

Instead of failing the evaluation, the comparison or filter handles the non-existing value as `null`. The comparison may return `false` or `null`. The filter doesn't include the list item in the result.

Since the engine handles the case gracefully, no warning should be printed. 

```
(null = 3) = false

(null < 3) = null

// 10.3.2.5 - example 3
[ { x: 1, y: 2 }, { x: null, y: 3 } ][ x < 2 ] = [ { x: 1, y: 2 } ]

// non-member matching
[ { x: 1 } ][ y > 1 ] = [ ]

[ { x: 1 }, { y: 2 } ][ y > 1 ] = [ { y: 2 } ]
```

**Environment**
* FEEL engine version: `1.15.2`
* Affects: 
  * Camunda Automation Platform 7: [7.x] 
  * Zeebe broker: [0.x]

saig0:
Related to the issue https://github.com/dmn-tck/tck/issues/536 that points to the following section of the DMN spec:

![image](https://user-images.githubusercontent.com/4305769/213127796-854fc211-3c3e-458a-9640-fd08336d88a6.png)

cjcalmeida:
I have a question about this topic:

[...]

PR #676, opened 2023-07-06
Handle non-existing variable in comparison

remcowesterhoud:
## Description

This PR is a WIP. A bunch of tests are still failing

Before when we encountered a non-existing variable in a comparison it would result in an error: "no variable found for name 'x'".
This commit changes this so we can deal with non-existing variables. They now get treated as ValNull and the comparison succeeds without errors.

## Related issues

Fixes part of #582

### DIFF
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
@@ -264,22 +264,22 @@ class InterpreterExpressionTest
 
   "A variable name" should "not be a key-word" in {
 
-    eval("some = true") shouldBe a[ValError]
-    eval("every = true") shouldBe a[ValError]
-    eval("if = true") shouldBe a[ValError]
-    eval("then = true") shouldBe a[ValError]
-    eval("else = true") shouldBe a[ValError]
-    eval("function = true") shouldBe a[ValError]
-    eval("for = true") shouldBe a[ValError]
-    eval("between = true") shouldBe a[ValError]
-    eval("instance = true") shouldBe a[ValError]
-    eval("of = true") shouldBe a[ValError]
-    eval("not = true") shouldBe a[ValError]
-    eval("in = true") shouldBe a[ValError]
-    eval("satisfies = true") shouldBe a[ValError]
-    eval("and = true") shouldBe a[ValError]
-    eval("or = true") shouldBe a[ValError]
-    eval("return = true") shouldBe a[ValError]
+    eval("some = true", Map("some" -> 1)) shouldBe a[ValError]
+    eval("every = true", Map("every" -> 1)) shouldBe a[ValError]
+    eval("if = true", Map("if" -> 1)) shouldBe a[ValError]
+    eval("then = true", Map("then" -> 1)) shouldBe a[ValError]
+    eval("else = true", Map("else" -> 1)) shouldBe a[ValError]
+    eval("function = true", Map("function" -> 1)) shouldBe a[ValError]
+    eval("for = true", Map("for" -> 1)) shouldBe a[ValError]
+    eval("between = true", Map("between" -> 1)) shouldBe a[ValError]
+    eval("instance = true", Map("instance" -> 1)) shouldBe a[ValError]
+    eval("of = true", Map("of" -> 1)) shouldBe a[ValError]
+    eval("not = true", Map("not" -> 1)) shouldBe a[ValError]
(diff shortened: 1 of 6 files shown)


src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
```diff
@@ -264,22 +264,22 @@ class InterpreterExpressionTest
...
+    eval("for = true", Map("for" -> 1)) shouldBe a[ValError]
+    eval("between = true", Map("between" -> 1)) shouldBe a[ValError]
+    eval("instance = true", Map("instance" -> 1)) shouldBe a[ValError]
+    eval("of = true", Map("of" -> 1)) shouldBe a[ValError]
+    eval("not = true", Map("not" -> 1)) shouldBe a[ValError]
+    eval("in = true", Map("in" -> 1)) shouldBe a[ValError]
+    eval("satisfies = true", Map("satisfies" -> 1)) shouldBe a[ValError]
+    eval("and = true", Map("and" -> 1)) shouldBe a[ValError]
+    eval("or = true", Map("or" -> 1)) shouldBe a[ValError]
+    eval("return = true", Map("return" -> 1)) shouldBe a[ValError]
```
remcowesterhoud, line 282: @saig0 I'm not quire sure about these tests. I think they weren't good tests to begin with. I'd expect an expression using a keyword as variable would fail during parsing, but apparently that's not the case.

From my limit understanding before this would also return a `ValError`, but it must've been an error about how the variable didn't exist.

The way it is not I make sure the variable is defined. This still returns a `ValError` as it the name is a keyword. It's a bit odd though, we don't care about it being a keyword, unless it's actually defined. If it's not defined we will consider it to be valid and consider it to be a `ValNull`.

saig0: I see. You're right. :+1:  The test case is partially broken. Some but not all expressions fail because of the parsing. The parser doesn't list all of the keywords as reserved words. 

The intention of the test is to verify that "a variable name should not be a keyword". The changes don't solve the issue.

I suggest making the expected error more clear. For example, by using a custom [matcher](https://www.scalatest.org/scaladoc/3.2.16/org/scalatest/matchers/BeMatcher.html):

```
  "A variable name" should "not be a key-word" in {
    eval("else") shouldBe aParseError
    // ...
  }

  class EvaluationErrorMatcher(expectedMessage: String) extends BeMatcher[Val] {
    override def apply(result: Val): MatchResult =
      result match {
        case ValError(failure) => MatchResult(
          failure.startsWith(expectedMessage),
          s"$result doesn't start with '$expectedMessage'",
          s"$result starts with '$expectedMessage'",
        )
        case _ => MatchResult(
          false,
          s"$result is not an error",
          s"$result is an error"
        )
      }
  }

  def anError(expectedMessage: String) = new EvaluationErrorMatcher(expectedMessage)
  def aParseError(): EvaluationErrorMatcher = anError(expectedMessage = "failed to parse expression")
```

And we should ignore the cases that are allowed (for now). 

---

:fast_forward: Feel free to fix this test case in a separate PR.

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterNumberExpressionTest.scala
```diff
@@ -180,11 +180,13 @@ class InterpreterNumberExpressionTest
...
-    eval("null < 2") should be(ValBoolean(false))
+    eval("2 > null") should be(ValNull)
+    eval("null < 2") should be(ValNull)
+  }
 
-    eval("null in < 2") should be(ValBoolean(false))
-    eval("null in (2..4)") should be(ValBoolean(false))
+  ignore should "compare with 'null in'" in {
+    eval("null in < 2") should be(ValNull)
+    eval("null in (2..4)") should be(ValNull)
```
saig0, line 189: :fast_forward: Let's create a follow-up issue to fix these cases as well as the unary-tests. :sweat_smile:

remcowesterhoud: https://github.com/camunda/feel-scala/issues/679

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
remcowesterhoud: https://github.com/camunda/feel-scala/pull/680 🙂

PR #680, opened 2023-07-12
Fix variable name should not be keyword test

remcowesterhoud:
## Description

This test should verify that a variable should not be allowed to be a key word. This should fail at parse time.
As we only checked if it returned an ValError this all succeeded. However, this is not accurate as the error wasn't always thrown at parse-time.

What would happen is that the expression got evaluated. It turned out the variable didn't exist and also throw the ValError. This means the assertion would succeed, but the test would actually be a false-positive.

With this custom matcher we can verify that we actually get the error we expect, by verifying if it's a parse error or not.
The keywords that have not been listed are now added to an ignore test case.

I tried adding the missing key words to the list and that broke a lot of tests. I figured there was a reason why they weren't there 😄 

## Related issues

I didn't create one, but it was mentioned [here](https://github.com/camunda/feel-scala/pull/676#discussion_r1258134087)


saig0 (approved):
@remcowesterhoud very nice. :cake:

PR #683, opened 2023-07-13
Remove the project-new-issues workflow

remcowesterhoud:
## Description

This workflow would add new issues to the ZPA project board. It was broken and would cause notifications of failed CI on every issue that's opened.

Besides this, not all FEEL issues are relevant to Zeebe. It doesn't make sense to add all these issues to the ZPA board. We can add the relevant ones manually when we create a new issue.

## Related issues

N/A


saig0 (approved):
@remcowesterhoud thank you. :rocket:

PR #685, opened 2023-07-13
Improve null handling in filter expressions

### LINKED ISSUE #582: Handle non-existing variables and context entries gracefully in comparison and filter expressions

korthout:
## Description

This changes the behavior of filter expressions to be more lenient.

Filter expressions now filter the list even when the filter expression is not evaluating to a boolean (for item comparison) or a number (for direct access).

When the filter expression would previously cause an error, the error is now suppressed and a list with only those values that match the filter expression is returned.

Please note that this pull request does not fully resolve issue #582. It does not yet deal with lists of contexts where the filter expression refers to missing properties of those contexts. This should be resolved in another pull request.

## Related issues

partially deals with #582 (filters over context lists are not yet covered)

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -789,9 +789,9 @@ class FeelInterpreter {
     val withBooleanFilter = (list: List[Val]) => mapEither[Val, Val](
       list,
       item =>
-        withBoolean(filter(item), {
-          case true => item
-          case false => conditionNotFulfilled
+        (filter(item) match {
+          case ValBoolean(true) => item
+          case _ => conditionNotFulfilled
         }).toEither,
       items => ValList(items.filterNot(_ == conditionNotFulfilled))
     )
@@ -809,7 +809,10 @@ class FeelInterpreter {
           case fulFilledItems: ValList => fulFilledItems
           case error => error
         }
-        case other => error(s"Expected boolean filter or number but found '$other'")
+        case _ => withBooleanFilter(list.tail) match {
+          case ValList(fulFilledItems) => ValList(fulFilledItems)
+          case error => error
+        }
       })
     ).getOrElse(
       // Return always an empty list if the given list is empty. Note that we would return `null`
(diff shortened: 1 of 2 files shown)


korthout:
@saig0 I've tried to describe what we did. Please have a look 🙇

saig0 (approved):
@korthout LTGM. :rocket:

korthout:
@saig0 Please have another look. I had to resolve a conflict in 04c123a.

saig0:
@korthout super. Thank you. :rocket:

Issue #654, opened 2023-06-02
Update mvn-scalafmt Maven plugin


saig0:
## Desciption 

We should update the `mvn-scalafmt` plugin that formats the Scala code. The updates were blocked because the new versions require Maven in version `3.6+` which is not available in our Jenkins CI (see [here](https://github.com/camunda/feel-scala/pull/127#issuecomment-639434623)).

At the moment, it seems that the code is not formatted anymore (maybe because of multiple source directories). 

requires #611

saig0:
Additionally, we should check on how to define formatting rules locally for the IDE.

PR #687, opened 2023-07-17
feat: Report suppressed failures

### LINKED ISSUE #260: Customize the error handling

saig0:
## Description

* Create a new API to report suppressed failures
* Mark the existing engine API as deprecated 
* Report a few suppressed failures, for example, for a missing variable. Other failures can be added later step-by-step. 

## Related issues

closes #260


src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -18,101 +18,11 @@ package org.camunda.feel.impl.interpreter
...
-  logger
-}
+import org.camunda.feel.syntaxtree.{Addition, ArithmeticNegation, AtLeastOne, ClosedConstRangeBoundary, ClosedRangeBoundary, Comparison, Conjunction, ConstBool, ConstContext, ConstDate, ConstDateTime, ConstDayTimeDuration, ConstInputValue, ConstList, ConstLocalDateTime, ConstLocalTime, ConstNull, ConstNumber, ConstRange, ConstRangeBoundary, ConstString, ConstTime, ConstYearMonthDuration, Disjunction, Division, Equal, EveryItem, Exp, Exponentiation, Filter, For, FunctionDefinition, FunctionInvocation, FunctionParameters, GreaterOrEqual, GreaterThan, If, In, InputEqualTo, InputGreaterOrEqual, InputGreaterThan, InputInRange, InputLessOrEqual, InputLessThan, InstanceOf, IterationContext, JavaFunctionInvocation, LessOrEqual, LessThan, Multiplication, NamedFunctionParameters, Not, OpenConstRangeBoundary, OpenRangeBoundary, PathExpression, PositionalFunctionParameters, QualifiedFunctionInvocation, RangeBoundary, Ref, SomeItem, Subtraction, UnaryTestExpression, Val, ValBoolean, ValContext, ValDate, ValDateTime, ValDayTimeDuration, ValError, ValFunction, ValList, ValLocalDateTime, ValLocalTime, ValNull, ValNumber, ValRange, ValString, ValTime, ValYearMonthDuration, ZonedTime}
```
remcowesterhoud, line 24: Just a casual 1186 character line 😆

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -288,7 +203,12 @@ class FeelInterpreter {
       // functions
       case FunctionInvocation(name, params) =>
         withFunction(findFunction(context, name, params),
-                     f => invokeFunction(f, params))
+          f => invokeFunction(f, params) match {
+            case ValError(failure) =>
+              error(EvaluationFailureType.FUNCTION_INVOCATION_FAILURE, s"Failed to invoke function '$name': $failure")
+              ValNull
```
remcowesterhoud, line 209: ❓ Isn't this a behavioural change? I get that we're in the process of changing errors to nulls. I just didn't expect it to be part of this PR.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -342,73 +262,71 @@ class FeelInterpreter {
...
     case _             => ValNull
   }
 
-  private def withBooleanOrFalse(x: Val, f: Boolean => Val): Val = x match {
+  private def withBooleanOrFalse(x: Val, f: Boolean => Val)(implicit context: EvalContext): Val = x match {
     case ValBoolean(x) => f(x)
-    case _ => {
-      logger.warn(s"Suppressed failure: expected Boolean but found '$x'")
+    case _ =>
+      error(EvaluationFailureType.INVALID_TYPE, s"Expected Boolean but found '$x'")
```
remcowesterhoud, line 323: ❌ I don't think this should be a failure. The method is called `withBooleanOrFalse` so there already is an expectation that `x` could be of a different type.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -676,14 +597,13 @@ class FeelInterpreter {
         case ValYearMonthDuration(y) => ValLocalDateTime(x.plus(y))
         case ValDayTimeDuration(y)   => ValLocalDateTime(x.plus(y))
         case _ =>
-          error(y, s"expect Year-Month-/Day-Time-Duration but found '$x'")
+          error(s"expect Year-Month-/Day-Time-Duration but found '$x'")
```
remcowesterhoud, line 538: 🔧 Wouldn't this be an `INVALID_TYPE` instead of `UNKNOWN`? There is a lot of these cases where the message says I expect one of these types, but I found something else. I'm pretty sure these are impossible to reach cases, but if it would occur it would be an invalid type.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -814,7 +725,7 @@ class FeelInterpreter {
     case ValDayTimeDuration(y) if (!y.isZero) =>
       withDayTimeDuration(x, x => ValNumber(x.toMillis / y.toMillis))
 
-    case _ => ValError(s"'$x / $y' is not allowed")
+    case _ => error(s"'$x / $y' is not allowed")
```
remcowesterhoud, line 666: ❓ Does it make sense to introduce some sort of `INVALID_ARITHMETIC` type?

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -870,11 +782,7 @@ class FeelInterpreter {
     }
 
     function.invoke(paramList) match {
-      case ValError(failure) => {
-        // TODO (saig0): customize error handling (#260)
-        logger.warn(s"Failed to invoke function: $failure")
-        ValNull
-      }
+      case e: ValError => e
```
remcowesterhoud, line 723: 🤔 This might be the answer to the behavioural change I mentioned earlier

remcowesterhoud (changes requested):
Great work @saig0! But much changes == much review comments 😄 Please have a look at them 👀

src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
```diff
@@ -64,7 +66,8 @@ class EvalContext(val valueMapper: ValueMapper,
   def merge(otherContext: EvalContext): EvalContext = new EvalContext(
     valueMapper = valueMapper,
     variableProvider = mergeVariableProvider(variableProvider, otherContext.variableProvider),
-    functionProvider = mergeFunctionProviders(functionProvider, otherContext.functionProvider)
+    functionProvider = mergeFunctionProviders(functionProvider, otherContext.functionProvider),
+    failureCollector = failureCollector
```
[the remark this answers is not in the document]
saig0: Currently, the other context should not have failures. In most cases, the new context is created with the same failure collector. 

If we see any need to merge the failures, we should do it. Until then, we can keep it simple. :smile:

src/main/scala/org/camunda/feel/impl/interpreter/EvaluationFailureCollector.scala
```diff
@@ -0,0 +1,21 @@
...
+  var failures: List[EvaluationFailure] = List.empty
+
+  def addFailure(failureType: EvaluationFailureType, failureMessage: String): Unit = {
+    failures :+= EvaluationFailure(
+      failureType = failureType,
+      failureMessage = failureMessage
+    )
+  }
+
+  def hasFailures: Boolean = failures.nonEmpty
```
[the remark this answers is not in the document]
saig0: Good point. Let's remove it. :wastebasket:

src/main/scala/org/camunda/feel/api/ParseResult.scala
```diff
@@ -0,0 +1,49 @@
...
+    */
+  def toEither: Either[Failure, ParsedExpression] =
+    if (isSuccess) Right(parsedExpression)
+    else Left(failure)
+
+}
+
+case class SuccessfulParseResult(parsedExpression: ParsedExpression) extends ParseResult {
+  override val isSuccess: Boolean = true
+  override val failure: Failure = Failure("<success>")
```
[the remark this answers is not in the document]
saig0: :bulb: This is inspired by the "Null object" pattern.

src/main/scala/org/camunda/feel/FeelEngine.scala
```diff
@@ -54,6 +48,9 @@ object FeelEngine {
 
   case class Failure(message: String)
 
+  /**
+    * @deprecated Use [[org.camunda.feel.api.FeelEngineBuilder]] instead.
```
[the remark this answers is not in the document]
saig0: Good point. A comment is not enough. :laughing:

src/test/scala/org/camunda/feel/impl/SuppressedFailuresTest.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  it should "report a suppressed failure for a non-existing function (with position arguments)" in {
+    engine.evaluateExpression("f(1, 2)") should reportFailure(
+      failureType = EvaluationFailureType.NO_FUNCTION_FOUND,
+      failureMessage = "no function found with name 'f' and 2 parameters"
+    )
+  }
+
+  it should "report a suppressed failure for a non-existing function (with named arguments)" in {
+    engine.evaluateExpression("f(x: 1, y: 2)") should reportFailure(
```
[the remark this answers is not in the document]
saig0: Yes. The failure message would be the same. 

If we improve the failure message and list the available functions, we can extend the test case.

src/test/scala/org/camunda/feel/impl/SuppressedFailuresTest.scala
```diff
@@ -0,0 +1,115 @@
+package org.camunda.feel.impl
+
+import org.camunda.feel.api._
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.matchers.{MatchResult, Matcher}
+
+class SuppressedFailuresTest extends AnyFlatSpec
```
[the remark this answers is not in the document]
saig0: The `UNKNOWN` is the default case. I plan to replace the default cases with a concrete failure type. And finally, remove the `UNKNOWN` case.

So, I would not write any specific test for it. It would be removed eventually. :wink:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Yes, currently the formatting doesn't work. :sweat_smile: 

I will improve the situation with https://github.com/camunda/feel-scala/issues/654.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: You're right. :+1: 

I could remove it from the PR. However, since we are planning to do this change anyway, I will just leave it here. :shushing_face:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Good catch. :+1: 

The function `withBooleanOrFalse()` is used in the `if-then-else` evaluation. If the condition is not a boolean, it is useful to report a failure. I will do the same for the function `withBooleanOrNull()`.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Yes, correct. I postponed these cases for later to limit the scope of this PR.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Yes, we could introduce this new failure type here. Let's do this later. :rocket:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Yes. Now, the error is propagated to the caller of `invokeFunction()` to report an error that contains also the name of the function.

saig0:
@remcowesterhoud thank you for the detailed review. :medal_sports: 

I addressed all your comments. Please have another look. :cake:

remcowesterhoud (approved):
Thanks for the changes, all looks good to me 🏆

PR #689, opened 2023-07-18
fix: Unary-test expression with null value

### LINKED ISSUE #679: Null in unary-test expressions should resolve to `ValNull`
**Describe the bug**
A comparison using null or a non-existing variable should resolve to a `ValNull`, except when it comes to comparisons. E.g.:

```
// Non existing less than
non-existing < 10
    ValNull

// null greater than
null > 10
    ValNull
```

For unary-test expressions one would expect the same behavior:

```
// Expected
null in < 10
    ValNull
```

Instead, it will return a `ValBoolean(true)`:

```
// Actual
null in < 10
    ValBoolean(true)
```

Note that this is a regression introduced in #676. Before the return of these expressions would be `ValBoolean(false)`!
There are tests verifying these that are ignored for now. They must be unignored when this is fixed:
- [InterpreterNumberExpressionTest](https://github.com/camunda/feel-scala/blob/main/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterNumberExpressionTest.scala#L187-L190)
- [InterpreterUnaryTest](https://github.com/camunda/feel-scala/blob/main/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala#L136-L140)

**Environment**
* FEEL engine version: None affected yet. Only on the main branch

saig0:
## Description

Align the behavior of unary-tests with a comparison to `null`. In both cases, the engine returns `null`. 

For example, the evaluation of `< 3` with `null` as the input value returns `null`. Same as the comparison `null < 3` returns `null`. 

Fixing the behavior in the parser by introducing a special case for simple-positive-unary-test (i.e. comparison with the input value).

## Related issues

closes #679

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -356,21 +356,14 @@ class FeelInterpreter {
   }
 
   private def unaryOpDual(
-      x: Val,
-      y: Val,
-      c: (Val, Val, Val) => Boolean,
-      f: Boolean => Val)(implicit context: EvalContext): Val =
+                           x: Val,
+                           y: Val,
+                           c: (Val, Val, Val) => Boolean,
+                           f: Boolean => Val)(implicit context: EvalContext): Val =
     withVal(
-      input,
-      _ match {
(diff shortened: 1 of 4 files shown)


remcowesterhoud (approved):
LGTM! 🚀

PR #691, opened 2023-07-19
New function get or else()

### LINKED ISSUE #618: New function to deal with optional values
If a process has optional variables then the FEEL expressions (e.g. variable mappings) need to deal with optional/nullable variables/values. Otherwise, the expression may fail.

Currently, the best option is to use an `if-then-else` block with a null-check. 

```
if x != null then x else "default_value"

// or

if is defined(x) then x else "default_value"
```
But it is very cumbersome.

This issue is also relevant to the integration with forms. In a form, a field may be optional or shown conditionally. The process must deal with these optional form values.

There is an easy way to deal with optional/nullable values. For example, by adding a new built-in function:

```
// signature
get or default(value: Any, default: Any) 

// usage
get or default("this", "default") // -> "this"

get or default(null, "default")  // -> "default"
```

Alternative function names: `get or else()`

**Alternative solutions**

Instead of adding a built-in function, it would be handy to include this optional/nullable handling natively in expressions.

We have examples in other programming languages. For example:

Kotlin:

```kotlin
x ?: "default" 
```

JavaScript:

```js
x || "default"

x ?? "default"
```

**Related issues**

* Camunda Autormation Platform 7: 
* Zeebe broker: 
* Support: https://jira.camunda.com/browse/SUPPORT-16485

nicpuppa:
## Description

Create a new function `get or else()` in order to allow users to easily deal with optional value, this new fucntion accept two parameters:

- value -> the optional value to check
- default -> the default value to return in case the first parameter is null

## Related issues

closes #618

### DIFF
--- a/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
+++ b/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
@@ -55,3 +55,26 @@ is defined(x)
 is defined(x.y)
 // false - if no variable "x" exists or it doesn't have a property "y"
 ```
+
+## get or else(value, default)
+
+Return the provided value parameter if not `null`, otherwise return the default parameter
+
+**Function signature**
+
+```js
+get or else(value: Any, default: Any): Val
+```
+
+**Examples**
+
+```js
+get or default("this", "default")
+// "this"
+
+get or default(null, "default")
+// "default"
+
+get or default(null, null)     
+// null
+```
--- a/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
+++ b/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
@@ -58,12 +58,14 @@ is defined(x.y)
 
 ## get or else(value, default)
 
+<MarkerCamundaExtension></MarkerCamundaExtension>
+
 Return the provided value parameter if not `null`, otherwise return the default parameter
 
 **Function signature**
 
 ```js
-get or else(value: Any, default: Any): Val
+get or else(value: Any, default: Any): Any
 ```
 
 **Examples**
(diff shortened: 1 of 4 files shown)


docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
```diff
@@ -55,3 +55,26 @@ is defined(x)
 is defined(x.y)
 // false - if no variable "x" exists or it doesn't have a property "y"
 ```
+
+## get or else(value, default)
+
```
saig0, line 60: :x: We should mark this function as Camunda extension (i.e. not part of the DMN spec).

We can add the marker by the following snipped, similar to the function above. :wink: 

```suggestion

<MarkerCamundaExtension></MarkerCamundaExtension>

```

docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
```diff
@@ -55,3 +55,26 @@ is defined(x)
...
+Return the provided value parameter if not `null`, otherwise return the default parameter
+
+**Function signature**
+
+```js
+get or else(value: Any, default: Any): Val
```
saig0, line 66: :x: We should change the return type of the function to `Any`. 

`Any` is part of the DMN specification. `Val` is an implementation class of the engine. 

```suggestion
get or else(value: Any, default: Any): Any
```

saig0 (changes requested):
@nicpuppa a great first contribution. Thank you. :tada: 

I have two minor comments about the documentation. Please have a look. :cookie:

nicpuppa:
Hi @saig0, thank you for the review 🚀 Requested changes applyed 🎉

saig0 (approved):
@nicpuppa LGTM. :rocket:

PR #705, opened 2023-09-07
feat: Handle non-existing variables

### LINKED ISSUE #674: Handle non-existing variables and context entries gracefully
**Describe the bug**
If I access a non-existing variable then the expression fails. To avoid failure, I must do a null check first.

```
x
// failed to evaluate expression 'x': no variable found for name 'x'
```

The behavior is the same for path expressions. If I access a non-existing context entry, the expression fails.

```
{a: 1}.b
// failed to evaluate expression '{a: 1}.b': context contains no entry with key 'b'
```

Related to https://github.com/camunda/feel-scala/issues/540.
Related to https://github.com/camunda/feel-scala/issues/582.

**To Reproduce**
1. Evaluate the expression `x`
2. Verify that the expression fails to evaluate

**Expected behavior**
The expression should return `null` if a variable or a context entry doesn't exist.

```
x = null

{a: 1}.b = null

{a: 1}.b.c = null

null.b = null
```

If the path expression is applied to a filter and a context doesn't contain the entry, the resulting list should contain `null` for this item.

```
// non-member path filtering
[ { x: 1 }, { y: 2 } ].y = [ null, 2 ]

[ { x: 1 } ].y = [ null ]
```

The expected behavior is aligned with the DMN spec. It is verified by the TCK test cases:
* https://github.com/dmn-tck/tck/pull/541
* https://github.com/dmn-tck/tck/pull/540

**Environment**
* FEEL engine version: `1.16.0`
* Affects: 
  * Zeebe broker: [0.x]

saig0:
## Description

Change the behavior of the FEEL engine to replace a non-existing variable and a non-existing input value with `null`. Previously, the evaluation failed in these cases. 

A note to the reviewer, the actual code fix is very small. The major changes are refactorings in the test classes to adopt the new test style. 

## Related issues

closes #674

### DIFF
--- a/src/test/scala/org/camunda/feel/api/context/CustomContextTest.scala
+++ b/src/test/scala/org/camunda/feel/api/context/CustomContextTest.scala
@@ -18,30 +18,32 @@ package org.camunda.feel.api.context
 
 import org.camunda.feel.FeelEngine
 import org.camunda.feel.FeelEngine.{Failure, UnaryTests}
-import org.camunda.feel.context.{
-  CustomContext,
-  FunctionProvider,
-  VariableProvider
-}
+import org.camunda.feel.context.{CustomContext, FunctionProvider, VariableProvider}
 import org.camunda.feel.context.VariableProvider.StaticVariableProvider
+import org.camunda.feel.impl.{EvaluationResultMatchers, FeelEngineTest}
 import org.camunda.feel.syntaxtree._
 import org.scalatest.matchers.should.Matchers
 import org.scalatest.flatspec.AnyFlatSpec
 
-class CustomContextTest extends AnyFlatSpec with Matchers {
-
-  val engine = new FeelEngine
+class CustomContextTest extends AnyFlatSpec with Matchers with FeelEngineTest with EvaluationResultMatchers {
 
   "A default context" should "provide its members" in {
-    engine.evalExpression("a", variables = Map("a" -> 2)) should be(Right(2))
-    engine.evalUnaryTests(
-      "2",
-      variables = Map(UnaryTests.defaultInputVariable -> 2)) should be(
-      Right(true))
+    evaluateExpression(
+      expression = "a",
+      variables = Map("a" -> 2)
+    ) should returnResult(2)
+
+    evaluateUnaryTests(
+      expression = "2",
+      inputValue = 2
+    ) should returnResult(true)
   }
 
   it should "return null if the variable doesn't exist" in {
-    engine.evalExpression("b", variables = Map("a" -> 2)) shouldBe Right(null)
+    evaluateExpression(
+      expression = "b",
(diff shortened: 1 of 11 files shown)


src/test/scala/org/camunda/feel/impl/interpreter/InterpreterNonExistingVariableExpressionTest.scala
```diff
@@ -16,82 +16,124 @@
...
+        failureType = EvaluationFailureType.NO_VARIABLE_FOUND,
+        failureMessage = "No variable found with name 'non_existing'")
+      )
+  }
+
+  "A non-existing input value" should "be equal to null" in {
+    // use parse + evaluate to avoid setting an input value
+    val parsedExpression = engine.parseUnaryTests("null").parsedExpression
+
+    engine.evaluate(parsedExpression) should (
```
nicpuppa, line 112: 💭 What about creating a new `evaluateParsedExpression` helper method in `FeelEngineTest`. So we can align all the test class with the new test engine. Improving also the readability

src/test/scala/org/camunda/feel/api/context/CustomContextTest.scala
```diff
@@ -18,65 +18,63 @@ package org.camunda.feel.api.context
...
+
+    engine.evaluateExpression(
+      expression = "floor(3.8)",
+      context = myCustomContext) should returnResult(3)
 
-    engine.evalExpression("b", context) should be
-    Left(
-      Failure(
-        "failed to evaluate expression 'b': no variable found for name 'b'"))
+    engine.evaluateUnaryTests(
```
nicpuppa, line 74: 💭 Maybe we can modify a bit the helper method `evaluateUnaryTests` by providing a custom context. So we can use it here

src/test/scala/org/camunda/feel/api/context/CustomContextTest.scala
```diff
@@ -18,65 +18,63 @@ package org.camunda.feel.api.context
...
-    engine.evalExpression("floor(3.8)", myCustomContext) should be(Right(3))
-    engine.evalUnaryTests("2", myCustomContext) should be(Right(true))
-  }
 
-  it should "fail on access to missing member" in {
-    val context = new CustomContext {
-      override def variableProvider: VariableProvider =
-        VariableProvider.StaticVariableProvider(Map.empty)
-    }
+    engine.evaluateExpression(
```
nicpuppa, line 65: 💭 What about creating a new `evaluateExpressionWithContext` helper method in `FeelEngineTest`. So we can align all the test class with the new test engine. Improving also the readability

nicpuppa (approved):
I love the refactor of the tests ❤️ 

LGTM 💪 🚀 

I just leave some thoughts, please have a look if you want 👀

PR #711, opened 2023-09-14
feat: Comparing different types returns null

### LINKED ISSUE #708: Comparing different types should return null
**Describe the bug**
If I compare two values of a different type, the engine returns a failure. But it should be more friendly and return `null` instead.  

This applies to expressions with the following operators: `=`, `!=`, `<`, `>`, etc.
And, this applies to unary-test expressions if the input value has a different type as the other value.

**To Reproduce**
1. Evaluate the following expression 
```
"foo" = 100
```
2. Verify that the engine returns a failure:
```
failed to evaluate expression '"foo" = 100': expected String but found 'ValNumber(100)'
``` 

**Expected behavior**
According to the DMN spec, the comparison of different types should return `null`.

Screenshot from 2023-09-08 14-33-47

(DMN 1.5, chapter 10.3.2.2, page 105)

- Related TCK test case: [0068-feel-equality](https://github.com/dmn-tck/tck/blob/master/TestCases/compliance-level-3/0068-feel-equality/0068-feel-equality.dmn#L233-L241)

**Environment**
* FEEL engine version: `1.16`
* Affects: 
  * Zeebe broker: [0.x]

saig0:
## Description

If an expression or a unary-test compares two values of different types then the evaluation returns null and reports a failure. Previously, the evaluation failed with an error.

The new behavior is aligned with the DMN specification and helps users to handle these cases.

## Related issues

closes #708

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -295,6 +295,7 @@ class FeelInterpreter {
           ValNull
         case i if (i.getClass != x.getClass || i.getClass != y.getClass) =>
           error(EvaluationFailureType.NOT_COMPARABLE, s"Can't compare $input with $x and $y")
+          ValNull
         case i => f(c(i, x, y))
       }
     )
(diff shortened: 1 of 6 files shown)


saig0:
@nicpuppa please review this new PR in the next few days. It is again about being more full-friendly - this time when comparing values of different types.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -295,6 +295,7 @@ class FeelInterpreter {
           ValNull
         case i if (i.getClass != x.getClass || i.getClass != y.getClass) =>
           error(EvaluationFailureType.NOT_COMPARABLE, s"Can't compare $input with $x and $y")
+          ValNull
```
nicpuppa, line 298: 🔧 Can we merge this two if statement ?

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -530,6 +534,7 @@ class FeelInterpreter {
         ValNull
       case _ if (x.getClass != y.getClass) =>
         error(EvaluationFailureType.NOT_COMPARABLE, s"Can't compare $x with $y")
+        ValNull
```
nicpuppa, line 537: 🔧 Can we merge this two if statement ?

src/test/scala/org/camunda/feel/impl/interpreter/ComparisonTypeTest.scala
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  it should "compare two null values" in {
+    evaluateExpression("null = null") should returnResult(true)
+  }
+
+  it should "return null if the values have a different type" in {
+    evaluateExpression("1 = true") should (returnNull() and reportFailure(
+      failureType = EvaluationFailureType.NOT_COMPARABLE,
+      failureMessage = "Can't compare ValNumber(1) with ValBoolean(true)"
```
nicpuppa, line 53: 💭 `Val` object are only for internal usage. Maybe we can find a way to build the error message with the correct type, like `boolean`, `string`, `number` .....

nicpuppa (approved):
LGTM 🥳 🚀 

I left some comments below, please have a look and let me know your thoughts @saig0 💪

src/test/scala/org/camunda/feel/impl/interpreter/ComparisonTypeTest.scala
saig0: I'm completely on your side! I want to improve the failure messages as well and make them more readable. 

I'll ship the improvement in a follow-up PR to keep the changes small. :+1:

PR #712, opened 2023-09-14
fix: `is defined()` function returns false if the value is null 

### LINKED ISSUE #695: is defined() always returns true
**Describe the bug**
The FEEL engine contains a function [is defined()](https://camunda.github.io/feel-scala/docs/reference/builtin-functions/feel-built-in-functions-boolean/#is-definedvalue). It can be used to check if a variable or context entry/property exists.

The behavior of this function changed since version `1.17.0`.

```js
// === before 1.17.0 ===
is defined("a value")
-> true

is defined(null) 
-> true

is defined(non_existing_variable) 
-> false 

is defined({}.non_existing_key) 
-> false

// === since 1.17.0 ===
is defined("a value")
-> true

is defined(null) 
-> true

is defined(non_existing_variable) 
-> true 

is defined({}.non_existing_key) 
-> true
``` 

**To Reproduce**
1. Evaluate the expression `is defined(non_existing_variable)`
2. Verify that the evaluation returns `true`

**Expected behavior**
The function `is defined()` behaves the same as in version `1.16.0`

**Environment**
* FEEL engine version: `1.17.0` (not yet released - on main branch)
* Affects: 
  * Zeebe broker: [0.x]

saig0:
## Description

Change the behavior of the function `is defined()` to return `true` if the value is not `null`. Previously, the function returned `true` if the value was `null`. It returned `false` if the value was a non-existing variable or context entry (i.e. a ValError).

The changes in the null-handling of the FEEL engine to replace a non-existing variable or context entry with `null` made this function useless. It returned always `true`.

A complete fix of the function is not possible because of the new null-handling. However, this change in the semantics of the function reduces the impact on the users (i.e. the regression).

The function still works as before for non-existing variables and context entries. But it may break for variables and context entries with a `null` value. See [here](https://github.com/camunda/feel-scala/issues/695#issuecomment-1711563232) for details.

## Related issues

closes #695

### DIFF
--- a/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
+++ b/docs/docs/reference/builtin-functions/feel-built-in-functions-boolean.md
@@ -30,9 +30,8 @@ not(null)
 
 <MarkerCamundaExtension></MarkerCamundaExtension>
 
-Checks if a given value is defined. A value is defined if it exists, and it is an instance of one of the FEEL data types including `null`.
-
-The function can be used to check if a variable or a context entry (e.g. a property of a variable) exists. It allows differentiating between a `null` variable and a value that doesn't exist.
+Checks if a given value is not `null`. If the value is `null` then the function returns `false`. 
+Otherwise, the function returns `true`.
 
 **Function signature**
 
@@ -47,7 +46,7 @@ is defined(1)
 // true
 
 is defined(null)
-// true
+// false
 
 is defined(x)
 // false - if no variable "x" exists
@@ -55,6 +54,13 @@ is defined(x)
 is defined(x.y)
 // false - if no variable "x" exists or it doesn't have a property "y"
 ```
+
+:::caution Breaking change
+
+This function worked differently in previous versions. It returned `true` if the value was `null`.
+Since this version, the function returns `false` if the value is `null`. 
+
+:::
 
 ## get or else(value, default)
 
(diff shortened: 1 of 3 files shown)


saig0:
@korthout please review this small PR in the next few days. :cake: 

I would like to hear your opinion on the fix and if it fixes the regression sufficiently. Happy to chat about it.

abbasadel:
Hi @nicpuppa , maybe you can review this PR since Nico is a bit busy right now.

nicpuppa:
💭 As the new null-handling made this function useless, and every fix introduce a breaking change, is not possible to deprecate it ? 
💭  The new behaviour of this function is closer to a null check, so the name `is defined()` is no more semantically correct, `is null()` would be more correct imho.

nicpuppa (approved):
LGTM 🚀

saig0:
> 💭 As the new null-handling made this function useless, and every fix introduce a breaking change, is not possible to deprecate it ?

Yes, but it would have no effect. The function could be used in existing BPMNs or DMN. To avoid breaking these models with the new version, we can't remove the function.

> 💭 The new behaviour of this function is closer to a null check, so the name is defined() is no more semantically correct, is null() would be more correct imho.

I agree. :sweat_smile: However, we need to keep the existing name to avoid breaking user models. 

For the future, I would recommend using a regular null-check instead of the function.

```
is defined(x)
// vs.
x != null
```

PR #717, opened 2023-09-19
feat: Make failure messages more readable

### LINKED ISSUE #713: Make failure messages more readable
## Description

Since version 1.17.0, we report failure messages if the evaluation results in `null` unintentionally. These failure messages are available for the caller of the API as part of the evaluation result.

Currently, these failure messages contain some technical details, like type information of the value. We should improve the readability of these messages.

For example

```
// Before
Can't compare ValNumber(1) with ValBoolean(true)

// After
Can't compare '1' with 'true'
```

---

```
it should "return null if the values have a different type" in {
    evaluateExpression("1 = true") should (returnNull() and reportFailure(
      failureType = EvaluationFailureType.NOT_COMPARABLE,
      failureMessage = "Can't compare ValNumber(1) with ValBoolean(true)"
```

> 💭 `Val` objects are only for internal usage. Maybe we can find a way to build the error message with the correct type, like `boolean`, `string`, `number` .....

_Originally posted by @nicpuppa in https://github.com/camunda/feel-scala/pull/711#discussion_r1325885261_

saig0:
## Description

* Uniform the failure messages and add missing failure types.
* Copy the existing string representation logic of the `string()` function into the value types.

## Related issues

closes #713

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
@@ -194,31 +194,10 @@ object ConversionBuiltinFunctions {
   private def isValidDecimalSeparator(separator: String) =
     separator == "," || separator == "."
 
-  private lazy val dateTimeOffsetZoneIdPattern =
-    Pattern.compile("(.*)([+-]\\d{2}:\\d{2}|Z)(@.*)")
-
   private def stringFunction = builtinFunction(
     params = List("from"),
     invoke = {
-      case List(ValString(from))  => ValString(from)
-      case List(ValBoolean(from)) => ValString(from.toString)
-      case List(ValNumber(from))  => ValString(from.toString)
-      case List(ValDate(from))    => ValString(from.format(dateFormatter))
-      case List(ValLocalTime(from)) =>
-        ValString(from.format(localTimeFormatter))
-      case List(ValTime(from)) => ValString(from.format)
-      case List(ValLocalDateTime(from)) =>
-        ValString(from.format(localDateTimeFormatter))
-      case List(ValDateTime(from)) => {
-        val formattedDateTime = from.format(dateTimeFormatter)
-        // remove offset-id if zone-id is present
-        val dateTimeWithOffsetOrZoneId = dateTimeOffsetZoneIdPattern
-          .matcher(formattedDateTime)
-          .replaceAll("$1$3")
-        ValString(dateTimeWithOffsetOrZoneId)
-      }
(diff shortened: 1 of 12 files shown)


saig0:
@nicpuppa I improved the failure messages as a follow-up from our last PR. Please have a look. :cookie:

src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
```diff
@@ -194,31 +194,10 @@ object ConversionBuiltinFunctions {
...
-        val formattedDateTime = from.format(dateTimeFormatter)
-        // remove offset-id if zone-id is present
-        val dateTimeWithOffsetOrZoneId = dateTimeOffsetZoneIdPattern
-          .matcher(formattedDateTime)
-          .replaceAll("$1$3")
-        ValString(dateTimeWithOffsetOrZoneId)
-      }
-      case List(duration: ValYearMonthDuration) => ValString(duration.toString)
-      case List(duration: ValDayTimeDuration)   => ValString(duration.toString)
+      case List(from)  => ValString(from.toString)
```
nicpuppa, line 200: This is great 🚀

src/test/scala/org/camunda/feel/impl/SuppressedFailuresTest.scala
```diff
@@ -49,48 +49,48 @@ class SuppressedFailuresTest extends AnyFlatSpec
...
-      failureMessage = "Can't compare ValBoolean(true) with ValNumber(2)"
+      failureMessage = "Can't compare 'true' with '2'"
     )
   }
 
   it should "report a suppressed failure if an addition has incompatible values" in {
     evaluateExpression("2 + true") should reportFailure(
       failureType = EvaluationFailureType.INVALID_TYPE,
-      failureMessage = "Expected Number but found 'ValBoolean(true)'"
+      failureMessage = "Expected number but found 'true'"
```
nicpuppa, line 66: ❓ How this is possible ? 🪄 

```Number -> number```

nicpuppa (approved):
LGTM, very big improvement 💪

src/test/scala/org/camunda/feel/impl/SuppressedFailuresTest.scala
saig0: I aligned the failure messages and changed the type names to start with a lowercase. See [here](https://github.com/camunda/feel-scala/blob/269a60fab0c227988c589a5ef217b114bc8a8243/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala#L304). :rocket:

PR #729, opened 2023-09-21
chore(ci): disbale fail on error for the javadoc 

yanavasileva:
## Description

* ci: add javadoc
* disable failure on error for javadoc => the javadocs won't be built but it's only one class so we can use this as workaround until the javadoc failure is fixed

## Related issues

https://github.com/camunda/camunda-bpm-platform/issues/3690


saig0 (approved):
@yanavasileva :+1:

PR #734, opened 2023-09-26
fix: typos

lzgabel:
## Description

Correct the example of the `get or else()` function in the doc.


saig0 (approved):
@lzgabel thank you for fixing the docs. :tada: :+1: 

I just saw the same issue. :laughing:

PR #749, opened 2023-10-20
fix: Regression in string() function for string and null values

### LINKED ISSUE #747: string() function wraps a string into a string value
**Describe the bug**
If I invoke the [string()](https://docs.camunda.io/docs/next/components/modeler/feel/builtin-functions/feel-built-in-functions-conversion/#stringfrom) function with a string argument, the function wraps the string into another string. As a result, the string value is escaped.

This behavior is different compared to previous versions of the FEEL engine (i.e. a regression).

The wrapped string value is complicated to handle because I need to unwrap it before using it. 

**To Reproduce**
1. Evaluate the expression 
```
string("foo")
```
2. Verify that the result is a string wrapped inside a string 
```
"\"foo\""
```

**Expected behavior**
If I invoke the `string()` function with a string value, it should return the string value as it is (i.e. no wrapping, no escaping).   

```
string("foo")        => "foo"
```

Related DMN-TCK test case [0079-feel-string-function](https://github.com/camunda/dmn-tck/blob/6f330dace2320b1c138cd7d8ba7e036c40087ebf/TestCases/non-compliant/0079-feel-string-function/0079-feel-string-function.dmn#L55-L63). 

**Environment**
* Affects: 
  * Zeebe broker: `8.3.0`

### LINKED ISSUE #748: string() function returns a string for a null value
**Describe the bug**
If I invoke the [string()](https://docs.camunda.io/docs/next/components/modeler/feel/builtin-functions/feel-built-in-functions-conversion/#stringfrom) function with `null`, the function returns a string `"null"`.

This behavior is different compared to previous versions of the FEEL engine (i.e. a regression).

**To Reproduce**
1. Evaluate the expression 
```
string(null)
```
2. Verify that the result is a string 
```
"null"
```

**Expected behavior**
If I invoke the `string()` function with `null`, it should return `null` (i.e. a null value, not a string).   

Screenshot from 2023-10-20 13-24-55

(DMN 1.5, 10.3.4.1 Conversion functions, page 138)

Related DMN-TCK test case [0079-feel-string-function](https://github.com/camunda/dmn-tck/blob/6f330dace2320b1c138cd7d8ba7e036c40087ebf/TestCases/non-compliant/0079-feel-string-function/0079-feel-string-function.dmn#L45-L53). 

**Environment**
* Affects: 
  * Zeebe broker: `8.3.0`

saig0:
## Description

Fix a regression in the `string()` function when it is invoked for string or null values. 

## Related issues

closes #747 
closes #748

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
@@ -231,8 +231,10 @@ object ConversionBuiltinFunctions {
 
   private def stringFunction = builtinFunction(
     params = List("from"),
-    invoke = { case List(from) =>
-      ValString(from.toString)
+    invoke = {
+      case List(ValNull)         => ValNull
+      case List(from: ValString) => from
+      case List(from)            => ValString(from.toString)
     }
   )
 
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
@@ -184,6 +184,10 @@ class BuiltinConversionFunctionsTest extends AnyFlatSpec with Matchers with Feel
     eval(""" string(1.1) """) should be(ValString("1.1"))
   }
 
+  it should "convert a string" in {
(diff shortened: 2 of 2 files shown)


saig0:
@nicpuppa please have a look at this small bug fix. It corrects a regression in the current version. Thank you. :cookie:

nicpuppa (approved):
LGTM 💪 

Thanks for refactoring the tests 👍

PR #786, opened 2023-12-22
fix: Variable resolution for custom context

### LINKED ISSUE #765: A list property gets overridden by variable within a filter expression
**Describe the bug**

A filter expression can access the properties of a list item by the property name. In the following example, the list is filtered by the property `loanId`. The variable `id` is provided by the context. 

```
sum({"loans" : [
  {"loanId" : "AAA001", "amount" : 10},
  {"loanId" : "AAA002", "amount" : 20},
  {"loanId" : "AAA001", "amount" : 50}
]}.loans[loanId = id].amount)
```

Since version `1.16.2`, the FEEL engine evaluates the expression differently if the context contains a variable `loanId`. The property `loanId` gets overridden by the variable `loanId`. It seems to be an issue with the variable resolution. 

This issue happens only if the context is provided by a custom context (i.e. not a static context). This is the case with the integration in Zeebe. 

**To Reproduce**
Use the following test case of the FEEL engine to reproduce the issue:

```scala
  "A filter expression" should "access an item property" in {
    evaluateExpression(
      expression =
        """sum({"loans" : [
                           {"loanId" : "AAA001", "amount" : 10},
                           {"loanId" : "AAA002", "amount" : 20},
                           {"loanId" : "AAA001", "amount" : 50}
                         ]}.loans[loanId = id].amount)""",
      context = new MyContext                                 // --> this is failing
      // variables = Map("id" -> "AAA002", "loanId" -> "AAA002") --> this is working
    ) should returnResult(20)
  }

  class MyContext extends CustomContext {

    val vars = Map("id" -> "AAA002", "loanId" -> "AAA002")

    override def variableProvider: VariableProvider = new VariableProvider {
      override def getVariable(name: String): Option[Any] = vars.get(name)

      override def keys: Iterable[String] = vars.keys
    }
  }
```

**Expected behavior**
In a list filter expression, the properties of a list item are not overridden by the context.

**Environment**
* FEEL engine version: `1.16.2`
* Affects: 
  * Zeebe broker: `8.2.18` https://github.com/camunda/zeebe/issues/15321
  * Support: https://jira.camunda.com/browse/SUPPORT-19328

### LINKED ISSUE #774: A context entry gets overridden by variable within a context liternal
**Describe the bug**
A context literal can access the previous context entries by their context key. In the following example, the context accesses the entry `a`. A variable with the same name `a` is provided by the context. 

```
{
  a: 10,
  b: a * 10
}
```

Since version `1.16.2`, the FEEL engine evaluates the expression differently if the context contains a variable `a`. The context entry `a` gets overridden by the variable `a`. It seems to be an issue with the variable resolution. 

This issue happens only if the context is provided by a custom context (i.e. not a static context). This is the case with the integration in Camunda 7's DMN engine and Zeebe. 

**To Reproduce**
Use the following test case of the FEEL engine to reproduce the issue:

```scala
  "A context literal" should "access a previous entry" in {
    evaluateExpression(
      expression =
        "{a: 10, b: a * 10}",
      context = new MyContext                                 // --> this is failing
      // variables = Map("a" -> 0) --> this is working
    ) should returnResult(100)
  }

  class MyContext extends CustomContext {

    val vars = Map("a" -> 0)

    override def variableProvider: VariableProvider = new VariableProvider {
      override def getVariable(name: String): Option[Any] = vars.get(name)

      override def keys: Iterable[String] = vars.keys
    }
  }
```

**Expected behavior**
In a context literal, the previous context entries are not overridden by the context.

**Environment**
* FEEL engine version: `1.16.1`
* Affects: 
  * Camunda Automation Platform 7: `7.20.0` https://github.com/camunda/camunda-bpm-platform/issues/3990

saig0:
## Description

Fix the variable resolution if the variables are provided by a custom context. 

Currently, the local variables within a list filter or a context literal are overridden by variables from the custom context.  

## Related issues

closes #765 
closes #774

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/EvalContext.scala
@@ -175,7 +175,7 @@ object EvalContext {
       case (StaticVariableProvider(thisVariables), StaticVariableProvider(otherVariables)) =>
         StaticVariableProvider(thisVariables ++ otherVariables)
       case (thisProvider, otherProvider)                                                   =>
-        CompositeVariableProvider(List(thisProvider, otherProvider))
+        CompositeVariableProvider(List(otherProvider, thisProvider))
     }
   }
 
@@ -195,7 +195,7 @@ object EvalContext {
         StaticFunctionProvider(allFunctions)
       }
       case (thisProvider, otherProvider)                                                   =>
-        CompositeFunctionProvider(List(thisProvider, otherProvider))
+        CompositeFunctionProvider(List(otherProvider, thisProvider))
     }
   }
 
--- a/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
+++ b/src/main/scala/org/camunda/feel/impl/parser/FeelParser.scala
@@ -692,14 +692,14 @@ object FeelParser {
   // replace escaped character with the provided replacement
   private def translateEscapes(input: String): String = {
     val escapeMap = Map(
-      "\\b" -> "\b",
-      "\\t" -> "\t",
-      "\\n" -> "\n",
-      "\\f" -> "\f",
-      "\\r" -> "\r",
-      "\\\""  -> "\"",
-      "\\'" -> "'",
-      "\\s" -> " "
+      "\\b"  -> "\b",
+      "\\t"  -> "\t",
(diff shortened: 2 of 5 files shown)


nicpuppa (approved):
LGTM 🚀 🍪

PR #824, opened 2024-04-08
feat: I can interrupt an expression evaluation

### LINKED ISSUE #823: I can't interrupt an expression evaluation
The feel engine is currently stuck while executing long-running expressions because of the lack of "cooperation mode" ([ref](https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html)). It is not possible to interrupt the execution from outside the engine. 

Related forum discussion: https://forum.camunda.io/t/feelengines-cooperation-mode-while-performing-longrunning-and-resource-consuming-operations/51049

> We are using org.camunda.feel.FeelEngine instance to evaluate FEEL expressions and trying to find a proper way to “sandbox” this operation and give a dedicated timeout for each one, in case the user tries to execute a malicious expression.
We’ve came across the following expression as an example of such harmful snippet which can bring the instance down:
>
> count(for x in 1..(2 ** 16) return {"power": 2 ** x}) > 0

The most classical solution is to repeatedly check whether the current thread was interrupted and throw an appropriate exception

vicmosin:
## Description

This simple change allows long-running feel evaluation/parsing processes to be gracefully interrupted. Consumers can execute the FEEL engine parsing/evaluations on a separate thread and be able to cancel them interruptedly.

## Related issues

closes #823

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -40,6 +40,7 @@ import java.time.{Duration, Period}
 class FeelInterpreter {
 
   def eval(expression: Exp)(implicit context: EvalContext): Val = {
+    // Check if the current thread was interrupted, otherwise long-running evaluations can not be interrupted and fully block the thread
     if (Thread.interrupted()) {
       throw new InterruptedException()
     }
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -39,7 +39,11 @@ import java.time.{Duration, Period}
   */
 class FeelInterpreter {
 
-  def eval(expression: Exp)(implicit context: EvalContext): Val =
+  def eval(expression: Exp)(implicit context: EvalContext): Val = {
+    if (Thread.interrupted()) {
+      throw new InterruptedException()
+    }
+
     expression match {
 
       // literals
@@ -269,6 +273,7 @@ class FeelInterpreter {
       case exp => error(EvaluationFailureType.UNKNOWN, s"Unsupported expression '$exp'")
 
     }
+  }
 
   private def mapEither[T, R](
       it: Iterable[T],
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
@@ -1,0 +1,86 @@
+/*
+ * Copyright Camunda Services GmbH and/or licensed to Camunda Services GmbH
+ * under one or more contributor license agreements. See the NOTICE file
+ * distributed with this work for additional information regarding copyright
+ * ownership. Camunda licenses this file to you under the Apache License,
+ * Version 2.0; you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.camunda.feel.impl.interpreter
+
+import org.camunda.feel.impl.{EvaluationResultMatchers, FeelEngineTest}
+import org.scalatest.flatspec.AnyFlatSpec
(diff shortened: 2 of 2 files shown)


vicmosin:
The solution allows consumers to execute the feel-engine parsing/evaluations on a separate threads and be able to cancel them itnerruptedly

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -39,7 +39,11 @@ import java.time.{Duration, Period}
   */
 class FeelInterpreter {
 
-  def eval(expression: Exp)(implicit context: EvalContext): Val =
+  def eval(expression: Exp)(implicit context: EvalContext): Val = {
+    if (Thread.interrupted()) {
+      throw new InterruptedException()
+    }
```
saig0, line 45: :wrench: We should add a comment to explain why we are checking for the interruption. 

It adds context for the next devs who is looking at this code.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0, line 1: :x: We should add a test case to verify that the interruption is working. And, to make sure that is will not be removed accidentally in the future.

saig0 (changes requested):
@vicmosin your changes look good. :tada: I added a few comments we must address before merging. Please have a look.

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
vicmosin: @saig0 Done, but not sure it's the proper implementation, I am not that familiar with scala

vicmosin:
Hey @saig0 , once merged can we get it as part of next minor release, let's say 1.17.6 ?

vicmosin:
@saig0 anything else I should adjust?

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
```diff
@@ -0,0 +1,86 @@
...
+import concurrent.{ExecutionContext}
+
+/** @author
+  *   Victor Mosin
+  */
+class InterpreterInterruptionTest
+    extends AnyFlatSpec
+    with Matchers
+    with FeelEngineTest
+    with EvaluationResultMatchers {
```
saig0, line 34: :wrench: `EvaluationResultMatchers` is not used. We can remove it.

```suggestion
class InterpreterInterruptionTest extends AnyFlatSpec with Matchers with FeelEngineTest {
```

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
```diff
@@ -0,0 +1,86 @@
...
+          )
+        } catch {
+          case _: InterruptedException =>
+            countDownLatch.countDown()
+        }
+      }
+    }
+
+    thread.start()
+    Thread.sleep(1000) // Let evaluation start
```
saig0, line 55: :wrench: We could decrease the waiting time to speed up the test. We don't need to wait so long. 

```suggestion
    Thread.sleep(100) // Let evaluation start
```

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
```diff
@@ -0,0 +1,86 @@
...
+    thread.interrupt()
+    countDownLatch.await()
```
saig0, line 57: :wrench: We could use the return value of `await()` to verify that the thread was interrupted. 

```suggestion
    val wasInterrupted = countDownLatch.await(1, TimeUnit.SECONDS)

    wasInterrupted should be(true)
```

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterInterruptionTest.scala
```diff
@@ -0,0 +1,86 @@
...
+
+    assert(!threadDump.contains("FeelInterpreter"))
```
saig0, line 59: :wrench: Using a thread dump to verify the interruption is a nice idea. However, it is a bit too much for this test case. It is enough to verify that the thread was interrupted by checking the count-down-latch.

Let's remove this part to keep it easy.

saig0 (changes requested):
@vicmosin very nice. :rocket: I see some improvements for the test case. Please have a look. :cookie:

vicmosin:
@saig0 done

saig0 (approved):
@vicmosin great. :zap: LGTM. :tada:

vicmosin:
@saig0 when to wait for a release, approx obviously?

saig0:
- #830

saig0:
@vicmosin I will build a patch release for 1.17 tomorrow. :building_construction:

PR #826, opened 2024-04-15
feat: add trim() function for whitespace stripping

### LINKED ISSUE #815: Add Trim function for whitespace stripping
I have a process where user input is displayed to other users including an id for example "Do extra testing during integration", that id can then later be used by other users in commands like `/markFinished Do extra testing during integration`, this fails when either the initial id or the command invocation has extra whitespace that the user is not expecting.

I would like there to be a trim function similar to other [languages like C#](https://learn.microsoft.com/en-us/dotnet/api/system.string.trim?view=net-8.0) where whitespace before or after the string is removed but internal whitespace is preserved.

**Related issues**

* Camunda Autormation Platform 7: 
* Zeebe broker:

skayliu:
## Description

Add the following built-in function `trim()`

```
// examples
trim("hello world")    // --> "hello world"
trim("hello world  ")    // --> "hello world"
trim("  hello world")    // --> "hello world"
trim("  hello world  ")    // --> "hello world"
trim(" hello   world ")    // --> "hello   world"
```

## Related issues

closes #815

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
@@ -37,7 +37,8 @@ object StringBuiltinFunctions {
     "ends with"        -> List(endsWithFunction),
     "matches"          -> List(matchesFunction, matchesFunction3),
     "split"            -> List(splitFunction),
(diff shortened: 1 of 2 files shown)


saig0 (approved):
@skayliu nice. :sparkles: LGTM. :rocket: 

Thank you for your contribution. :tada:

PR #831, opened 2024-04-19
Release 1.17.6

saig0:
## Description

Released new patch version 1.17.6. :building_construction: 

## Related issues

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,7 +11,7 @@
 
   <groupId>org.camunda.feel</groupId>
   <artifactId>feel-engine</artifactId>
-  <version>1.17.6-SNAPSHOT</version>
+  <version>1.17.6</version>
   <name>FEEL Scala Engine</name>
 
   <properties>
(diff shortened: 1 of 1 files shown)


vicmosin:
@saig0 I think something went wrong.. Updating to recent camunda 7.21 and providing the feel-engine 1.17.6 brings the following incompatibility issue

Bildschirmfoto 2024-04-24 um 08 39 14

saig0:
@vicmosin thank you for reporting. :+1: This compiler issue seems to be related to the shaded JAR that bundles the Scala library. 

Can you share a repository to reproduce the issue?
Did it work with version 1.17.5?

vicmosin:
I have no repository for this yet, but basically the builtin 1.17.5 (which is part of camunda-engine-dmn:7.21.0) compiles without any issue, but once I exclude it and provide the 1.17.6 (with the same camunda-engine-dmn:7.21.0) it starts to complain

vicmosin:
@saig0 just making sure you got the last message :)

saig0:
@vicmosin sounds like a dependency issue. Make sure to use the same versions. Maybe, a different Scala library is used from somewhere. Since the DMN uses the bundled JAR with the Scala library, use the bundled Scala classes should be used.

PR #832, opened 2024-04-19
feat: Make the FeelEngineBuilder available from Java/Kotlin

### LINKED ISSUE #724: The FeelEngineBuilder is not usable from Java
**Describe the bug**
From a Java application, I can't use the new builder `FeelEngineBuilder`. The (default) constructor is not visible.  

Screenshot from 2023-09-20 13-56-00

**To Reproduce**
1. In a Java class, add the following code snippet:
```
new FeelEngineBuilder().build()
```
2. Verify that the statement doesn't compile 

**Expected behavior**
I can use the new FEEL engine builder from a Java application.

**Environment**
* Affects: 
  * Zeebe broker: [0.x]

vicmosin:
## Description

Refactored implementation

## Related issues

closes #724

### DIFF
--- a/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
+++ b/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
@@ -21,9 +21,6 @@ import org.camunda.feel.FeelEngine.{Configuration, defaultFunctionProvider}
 import org.camunda.feel.context.FunctionProvider
 import org.camunda.feel.valuemapper.{CustomValueMapper, ValueMapper}
 import org.camunda.feel.valuemapper.ValueMapper.CompositeValueMapper
-import scala.annotation.StaticAnnotation
-
-class javaStatic extends StaticAnnotation
 
 /** Builds a new instance of the FEEL engine. Use the setters to customize the engine.
   */
@@ -80,6 +77,5 @@ case class FeelEngineBuilder private (
 
 object FeelEngineBuilder {
 
-  @javaStatic
   def create(): FeelEngineBuilder = FeelEngineBuilder()
 }
--- a/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
+++ b/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
@@ -21,6 +21,9 @@ import org.camunda.feel.FeelEngine.{Configuration, defaultFunctionProvider}
 import org.camunda.feel.context.FunctionProvider
 import org.camunda.feel.valuemapper.{CustomValueMapper, ValueMapper}
 import org.camunda.feel.valuemapper.ValueMapper.CompositeValueMapper
+import scala.annotation.StaticAnnotation
+
+class javaStatic extends StaticAnnotation
 
 /** Builds a new instance of the FEEL engine. Use the setters to customize the engine.
   */
@@ -91,4 +94,10 @@ case class FeelEngineBuilder() {
     )
   )
 
+}
+
+object FeelEngineBuilder {
+
+  @javaStatic
+  def create(): FeelEngineBuilder = FeelEngineBuilder()
 }
--- a/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
+++ b/src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
@@ -24,40 +24,58 @@ import org.camunda.feel.valuemapper.ValueMapper.CompositeValueMapper
 
 /** Builds a new instance of the FEEL engine. Use the setters to customize the engine.
   */
-case class FeelEngineBuilder private (
-    functionProvider: FunctionProvider = defaultFunctionProvider,
-    valueMapper: ValueMapper = FeelEngine.defaultValueMapper,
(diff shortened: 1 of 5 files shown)


vicmosin:
Mh.. not sure what causes this:

```
[INFO] Comparing to version: 1.17.5
Error:  7002: org.camunda.feel.api.FeelEngineBuilder: Method 'public scala.Function1 curried()' has been removed
Error:  7002: org.camunda.feel.api.FeelEngineBuilder: Method 'public scala.Function1 tupled()' has been removed
```

src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
```diff
@@ -74,3 +77,9 @@ case class FeelEngineBuilder private (
   )
 
 }
+
+object FeelEngineBuilder {
+
+  @javaStatic
```
saig0, line 83: :thought_balloon: I don't see that the annotation is required. I can access the static method in the Java class without the annotation.

src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
```diff
@@ -74,3 +77,9 @@ case class FeelEngineBuilder private (
...
+  def create(): FeelEngineBuilder = FeelEngineBuilder()
```
saig0, line 84: :+1: Nice.

saig0: :wrench: Additionally, we could create a convenient method for Java applications. Usually, every Java application will use the JavaValueMapper. 

For example:

```
def forJava(): FeelEngineBuilder = FeelEngineBuilder().withCustomValueMapper(new JavaValueMapper)
```

src/test/java/org/camunda/feel/api/FeelEngineBuilderTest.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package org.camunda.feel.api;
+
+public class FeelEngineBuilderTest {
+
+    public void build() {
+        FeelEngineApi engine = FeelEngineBuilder.create().build();
+    }
+}
```
saig0, line 24: :x: We should write a real unit test. 

One option is to use the Java class as a Java application and write a Scala test that uses the Java application.

saig0 (changes requested):
@vicmosin thank you for your contribution. :tada: I added a few comments to improve the solution. Please have a look. :cookie:

saig0:
@vicmosin the build failures are reported by the Clirr plugin. By adding an object to the builder, it changes the Scala internal signature. However, this is okay. We need to exclude these changes in the plugin configuration [here](https://github.com/camunda/feel-scala/blob/main/pom.xml#L254-L349).

src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
vicmosin: Done

src/test/java/org/camunda/feel/api/FeelEngineBuilderTest.java
vicmosin: Not sure if you meant it this way, but I added a real test.

src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
vicmosin: @saig0 tbh I am not sure I really understood your idea about `forJava` method.. Could you please elaborate?

saig0:
@vicmosin do you want to continue contributing to this PR? The last update was one month ago.

vicmosin:
@saig0 I am sorry, I've been busy last month, will be available at the beginning of the june. If it's urgent, feel free to take over. Thnx

saig0:
@vicmosin thank you for the update. :+1: No need to rush. June is fine. :sun_with_face:

src/main/scala/org/camunda/feel/api/FeelEngineBuilder.scala
vicmosin: Done

src/test/scala/org/camunda/feel/api/FeelEngineBuilderTest.scala
```diff
@@ -0,0 +1,28 @@
...
+ */
+package org.camunda.feel.api
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class FeelEngineBuilderTest extends AnyFlatSpec with Matchers {
+
+  "A FeelEngineBuilder" should "successfully build a FeelEngineApi instance" in {
+    val engine: FeelEngineApi = FeelEngineBuilder.create().build()
```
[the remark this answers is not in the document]
vicmosin: Donr

saig0 (approved):
I see some minor improvements. However, let's merge this PR first. :rocket:

saig0:
@vicmosin no problem. :laughing: It's merged. :rocket:

PR #834, opened 2024-04-20
feat: add built-in function to generate a UUID string

### LINKED ISSUE #796: Add built-in function to generate an UUID
We use UUID's a lot in requests and match with them in replies, so being able to generate a UUID from within FEEL itself would be a huge benefit.

Alternatives are generating one with a Service Worker, or even requesting one from an online source with the REST connector, but these look like aweful alternatives.

As requested here: https://forum.camunda.io/t/built-in-function-for-uuid/35018/14

Being able to do `=UUID()` to create an input variable with an UUID.

```
// function signature
uuid(): String

// usage
uuid()    // -> "7793aab1-d761-4d38-916b-b7270e309894"
uuid()    // -> "df09d5c2-e9f0-432d-9027-aaa5db56b8b5"
```

**Related issues**

* https://github.com/camunda/zeebe/issues/10987

skayliu:
## Description

Add the following built-in function `uuid()`.

```
// function signature
uuid(): String

// usage
uuid()    // -> "7793aab1-d761-4d38-916b-b7270e309894"
uuid()    // -> "df09d5c2-e9f0-432d-9027-aaa5db56b8b5"
```
## Related issues

closes #796

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,6 +18,7 @@
     <scala.version>2.13.13</scala.version>
     <scala.binary.version>2.13.6</scala.binary.version>
     <version.log4j>2.23.1</version.log4j>
+    <version.uuid>5.0.0</version.uuid>
 
     <plugin.version.shade>3.5.2</plugin.version.shade>
     <plugin.version.gpg>1.6</plugin.version.gpg>
@@ -88,6 +89,12 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.fasterxml.uuid</groupId>
+      <artifactId>java-uuid-generator</artifactId>
+      <version>${version.uuid}</version>
     </dependency>
 
     <dependency>
--- a/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
@@ -25,6 +25,8 @@ import scala.util.Try
 
 object StringBuiltinFunctions {
 
+  private lazy val generator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())
+
   def functions = Map(
     "substring"        -> List(substringFunction, substringFunction3),
     "string length"    -> List(stringLengthFunction),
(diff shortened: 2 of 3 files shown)


pom.xml
```diff
@@ -90,6 +91,12 @@
       <artifactId>slf4j-api</artifactId>
     </dependency>
 
+    <dependency>
+      <groupId>com.fasterxml.uuid</groupId>
+      <artifactId>java-uuid-generator</artifactId>
+      <version>${version.uuid}</version>
+    </dependency>
```
saig0, line 98: :question: Why do you want to use this library?

Java has a built-in `UUID` class. You could generate UUID by using `UUID.randomUUID()`.

saig0 (commented):

The changes look very good. :+1: Before accepting it, I have a question about the new dependency. Please have a look.

pom.xml
skayliu: if you insist on the Java build-in `UUID.randomUUID()`, i can change to it.

saig0: @skayliu very interesting. :+1: I see that this library is also used by Camunda 7. :bulb: That is a good indication for using it here too.

src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
```diff
@@ -262,4 +264,13 @@ object StringBuiltinFunctions {
         ValString(string.trim)
       }
     )
+
+  private def uuidFunction =
+    builtinFunction(
+      params = List(),
+      invoke = { case List() =>
+        val uuid  = Generators.timeBasedGenerator.generate()
```
saig0, line 272: :wrench: In Camunda 7, we create only one instance of the generator, by including the ethernet address ([ref](https://github.com/camunda/camunda-bpm-platform/blob/master/engine/src/main/java/org/camunda/bpm/engine/impl/persistence/StrongUuidGenerator.java#L44C32-L44C94)). We could do the same and define the generate as lazy val field.

```
private lazy val generator = Generators.timeBasedGenerator(EthernetAddress.fromInterface())

// ...
invoke = { case List() => ValString(generator.generate.toString()) }
```

saig0 (commented):
@skayliu I have one remaining minor comment about the generator usage. Please have a look if you agree. :cookie:

src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
skayliu: @saig0, It's done.

saig0 (approved):
@skayliu LGTM. :tada:

Issue #825, opened 2024-04-12
Add function to parse an escaped JSON string 


skayliu:
**Is your feature request related to a problem? Please describe.**
When use the REST connector to call a service, i got the response like below.
Now i need to extract the `taskTodoList` into an array to the result expression.

```
{
    "status": 200,
    "headers": {
        "date": "Fri, 12 Apr 2024 01:29:49 GMT",
        "server": "nginx",
        "transfer-encoding": "chunked",
        "x-content-type-options": "nosniff",
        "x-xss-protection": "1; mode=block",
        "x-frame-options": "SAMEORIGIN",
        "content-type": "application/json;charset=utf-8",
        "connection": "keep-alive",
        "strict-transport-security": "max-age=172800; includeSubDomains"
    },
    "body": {
        "RESPONSE": {
            "RETURN_CODE": "S",
[...]

PR #847, opened 2024-05-06
feat: add built-in function to encode a string as base64

### LINKED ISSUE #846: Support encoding of strings to base64

[Some APIs](https://docs.github.com/de/rest/repos/contents?apiVersion=2022-11-28#create-or-update-file-contents) require data to be sent as base64 encoded strings. Even for text values like textfile contents. It would therefore be handy to be able to encode strings right inside the feel engine via a OOTB function that supports strings.

We should support an `encode base64` function that accepts a string as the input and returns a base64 encoded string as the output.

**Related issues**

https://github.com/camunda/connectors/issues/2023

sbuettner:
## Description

Add function to encode a string as base64

## Related issues

closes #846

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
@@ -44,7 +44,7 @@ object StringBuiltinFunctions {
     "extract"          -> List(extractFunction),
     "trim"             -> List(trimFunction),
     "uuid"             -> List(uuidFunction),
-    "encode base64"    -> List(encodeBase64Function)
+    "to base64"        -> List(toBase64Function)
   )
 
   private def substringFunction = builtinFunction(
@@ -277,11 +277,11 @@ object StringBuiltinFunctions {
       }
     )
 
-  private def encodeBase64Function =
+  private def toBase64Function =
     builtinFunction(
-      params = List("string"),
-      invoke = { case List(ValString(string)) =>
-        val bytes = string.getBytes(StandardCharsets.UTF_8)
+      params = List("value"),
+      invoke = { case List(ValString(value)) =>
+        val bytes = value.getBytes(StandardCharsets.UTF_8)
         ValString(Base64.getEncoder.encodeToString(bytes))
       }
     )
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
```diff
@@ -42,7 +43,8 @@ object StringBuiltinFunctions {
     "split"            -> List(splitFunction),
     "extract"          -> List(extractFunction),
     "trim"             -> List(trimFunction),
-    "uuid"             -> List(uuidFunction)
+    "uuid"             -> List(uuidFunction),
+    "encode base64"    -> List(encodeBase64Function)
```
saig0, line 47: :wrench: We could rename the function to `to base64()` to align with other data transformation functions, like `to json()` ([ref](https://github.com/camunda/feel-scala/issues/825)). 

What do you think about changing the name?

src/test/scala/org/camunda/feel/impl/builtin/BuiltinStringFunctionsTest.scala
```diff
@@ -173,4 +173,11 @@ class BuiltinStringFunctionsTest extends AnyFlatSpec with Matchers with FeelInte
 
     eval(" string length(uuid()) ") should be(ValNumber(36))
   }
+
+  it should "return a string encoded as base64" in {
```
saig0, line 177: :wrench: We need to group the test cases by name. By using the keyword `it`, it adds the test case to the previous group above. 

```suggestion
  "A encode base64() function" should "return a string encoded as base64" in {
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinStringFunctionsTest.scala
```diff
@@ -173,4 +173,11 @@ class BuiltinStringFunctionsTest extends AnyFlatSpec with Matchers with FeelInte
...
+
+    eval(""" encode base64("FEEL") """) should be(ValString("RkVFTA=="))
+
+    eval(""" encode base64("Camunda") """) should be(ValString("Q2FtdW5kYQ=="))
```
saig0, line 181: :wrench: We should add another test case that invokes the function with the named argument. This ensure that we don't change the parameter name accidentally. 

```
eval(""" encode base64(string: "Camunda") """) should be(ValString("Q2FtdW5kYQ=="))
```

saig0 (changes requested):
@sbuettner thank you for your contribution. :tada: 

Looks very good. :+1: I have a few minor comments. Please have a look. :cookie:

src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
sbuettner: Yes as it was just a proposal and I wasnt happy with it either.

sbuettner:
@saig0 Please take another look.

saig0 (approved):
@sbuettner LGTM. :rocket:

saig0:
@sbuettner when do you need this feature? 

The next version 1.18.0 is planned for August/September. However, we could build the version earlier if needed.

sbuettner:
Thanks for asking @saig0. Earlier would be great as we depend on it for some document related Connectors.

PR #853, opened 2024-05-23
fix: number() should return null instead of failing the evaluation

### LINKED ISSUE #731: number() should return null instead of failing the evaluation
**Describe the bug**

**To Reproduce**
1. Evaluate the expression 
```
number("x")
```

**Expected behavior**

**Environment**
* Affects: 
  * Zeebe broker: [0.x] 

skayliu:
## Description

The function `number()` should return `null` if the string is not a valid number.

## Related issues

closes #731

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
@@ -19,26 +19,11 @@ package org.camunda.feel.impl.builtin
 import org.camunda.feel.impl.builtin.BuiltinFunction.builtinFunction
 import org.camunda.feel.syntaxtree._
 import org.camunda.feel.valuemapper.ValueMapper
-import org.camunda.feel.{
-  Date,
-  YearMonthDuration,
-  isDayTimeDuration,
-  isLocalDateTime,
-  isOffsetDateTime,
-  isOffsetTime,
-  isValidDate,
-  isYearMonthDuration,
-  stringToDate,
-  stringToDateTime,
-  stringToDayTimeDuration,
-  stringToLocalDateTime,
-  stringToLocalTime,
-  stringToNumber,
-  stringToYearMonthDuration
-}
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
```diff
@@ -169,14 +154,17 @@ class ConversionBuiltinFunctions(valueMapper: ValueMapper) {
   private def numberFunction =
     builtinFunction(
       params = List("from"),
-      invoke = { case List(ValString(from)) =>
-        ValNumber(from)
+      invoke = {
+        case List(ValString(from)) if !isNumeric(from) => ValNull
```
saig0, line 158: :wrench: Instead of returning `ValNull`, we could return a `ValError()` with a proper failure message. As a result, the engine would use `null` but also produce a warning for debugging purposes.

We can verify the warning in the test case. See the example in [ComparisonTypeTest.scala](https://github.com/camunda/feel-scala/blob/main/src/test/scala/org/camunda/feel/impl/interpreter/ComparisonTypeTest.scala#L51-L55).

src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
```diff
@@ -187,6 +175,12 @@ class ConversionBuiltinFunctions(valueMapper: ValueMapper) {
   private def numberFunction3 = builtinFunction(
     params = List("from", "grouping separator", "decimal separator"),
     invoke = {
+      case List(ValString(from), ValString(grouping), ValString(decimal)) if !isNumeric(from)     =>
+        ValNull
+      case List(ValString(from), ValNull, ValString(decimal)) if !isNumeric(from)                 =>
+        ValNull
+      case List(ValString(from), ValString(grouping), ValNull) if !isNumeric(from)                =>
+        ValNull
```
saig0, line 183: :wrench: We could combine these cases into one. 

```suggestion
      case List(ValString(from), _, _) if !isNumeric(from)     =>
        ValNull
```

saig0 (changes requested):

I added a few comments to fix and improve the solution. Please have a look. :cookie:

saig0:
Close this PR is favor of #892.

Issue #856, opened 2024-05-28
The `context()` function doesn't work for a list with a value with a custom context


saig0:
**Describe the bug**
The `context()` function doesn't work for a list value that contains a custom context. This is the case when the function is invoked in Camunda 8 with a process variable as the argument. See the details in the issue: https://github.com/camunda/camunda/issues/12740.

The function works fine if it is invoked with a list literal.

**To Reproduce**
1. Add the following test to `BuiltinContextFunctionsTest` and run it
2. Verify that the test fails with the message: `the evaluation didn't returned 'Map(a -> 1)' but 'null'`  

<details>
  <summary>ScalaTest</summary>
  <pre>
it should "return a context with entries from a custom context" in {

    evaluateExpression(
      expression = """context(list)""",
      variables = Map(
        "list" -> List(
          ValContext(
            new MyCustomContext(
              Map(
                "key"   -> "a",
                "value" -> 1
              )
            )
          )
        )
      )
    ) should returnResult(Map("a" -> 1))
  }
</pre>
</details>

**Expected behavior**
I can invoke the `context()` function with a list that a custom context.

**Environment**
* FEEL engine version: 1.17
* Affects: 
  * Camunda Automation Platform 7: [7.x] 
  * Zeebe broker: https://github.com/camunda/camunda/issues/12740
  * Support case: https://jira.camunda.com/browse/SUPPORT-16886

PR #861, opened 2024-05-30
[Backport 1.16] fix: Invoke context() function with a list containing a custom context

saig0:
## Description

Backport of #858 to `1.16`.

## Related issues

relates to #856


berkaycanbc (approved):
Nice, we also used the new expression evaluation function in the existing tests. 🚀

PR #870, opened 2024-06-26
fix(maven-central): Use tokens to authenticate with Maven Central for…

andromaqui:
… publishing artifacts

## Description

This pr modifies the secrets used for autheticating to `oss.sonatype.org`. More specficially, Maven has enforced token based authentication and we therefore need to separate the credentials used for `oss.sonatype.org` and `s01.oss.sonatype.org`

## Related issues

related to https://github.com/camunda/team-infrastructure/issues/640


saig0 (approved):
:+1:

Issue #864, opened 2024-06-10
[Bug] Incorrect/inconsistent unary test evaluation when nulls are involved.


spearsandshields:
**Describe the bug**
There is some inconsistent handling of null when evaluating unary tests. 
I will try to demonstrate below using real examples and screenshots.

1. evaluating `null > 20`
    1. when input value is a random number, this evaluates to false. 
        
![image](https://github.com/camunda/feel-scala/assets/11656422/6bc48134-ba7d-43e9-b199-12dce34cdc91)
    2. when input value is null. I would expect the same (since the unary test is not a simple unary test and is not using the input value at all). However, we are getting true. null > 20 => true? This seemed hard to reason about.

![image](https://github.com/camunda/feel-scala/assets/11656422/9b0511da-4268-4df7-8450-6c4f3c24217a)
2. evaluating `? > 20`, and when input value is null. 
![image](https://github.com/camunda/feel-scala/assets/11656422/41fc210c-587d-4699-93ba-46b640d4f252)
    However, when i run in expression mode, or using feelin engine, I would see this evaluate to null. If I run it against dsntk, I would get null as the return value as well. 
![image](https://github.com/camunda/feel-scala/assets/11656422/b2ce7f34-84c4-4bb3-b57c-a2254ee6ba82)

3. Technically speaking DMN spec did not talk about the handling of null when inequality comparison, however based on the guidance over equality comparison, it seems reasonable to suggest that if the two values are not of the same kind, we should expect null as the return value.
![image](https://github.com/camunda/feel-scala/assets/11656422/390c5c0c-18e0-4c13-8f16-2bb37641d23e)

**To Reproduce**
1.Type in the values as shown in the screenshot in corresponding playground sites. or go to link to spot the inconsistency
https://camunda.github.io/feel-scala/docs/playground/?expression-type=unary-tests&expression=PyA%2BIDIw&context=ewp9&input-value=bnVsbA%3D%3D
https://nikku.github.io/feel-playground/?e=%3F+%3E+20&c=%7B%0A++%22%3F%22%3A+null%0A%7D&t=expression&st=true

**Expected behavior**
This is business critical as it directly impact which rule is determined as a match or not a match when nulls are involved in the evaluation context. It seems like the general consensus is to return null, or consider a rule as not a match for unary tests such as `? > 20`, when ? is null. 

Currently, these null resulting input values are causing unexpected behavior in rule match determination.  I think null will be a more reasonable/spec-compliant return value.
[...]

PR #882, opened 2024-08-01
fix: better null handling for unary tests involving special character ?

spearsandshields:
## Description

This PR is drafted to address null value handling in unary test, especially when special charactor ? is involved.

Gist of the fix:
Whenever something evaluates to null, due to errors. Such as not-comaparable data types, we need a way to embed/box that information, and pass it along to the callers. So the callers can determine how they want to handle these special values. 

Therefore, I created a artifact valueType called ErroredVal, this is different from ValError, as ValError results in not-resumable error. ErroredValue allow the logic to continue, and callers can decide when/how to handle it at their choosing. 
Adding this mechanism allow me to swap some ValNull to boxed ErroredVal. 

After this, we need to ensure the ErroredVal does not affect the final result. In order to achieve that, I moved logic from eval to evalInternal, and made eval function behave like a wrapper than returns the unboxed ErroredVal.  (this allow tests to pass as normal) 

With the additional of this new mechanism, I made some changes to unaryTestExpression:
1. add expression as additional method parameter. As we need to check the tokens and see if ? is used in the expression. 
2. when it is used, check for ErroredVal and handle them accordingly. 
 

* 

## Related issues
https://github.com/camunda/feel-scala/issues/864

closes #

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -766,6 +766,7 @@ class FeelInterpreter {
                 case _ if x.isInstanceOf[ValList] =>
                   ValBoolean(false) // the expression is a list but doesn't contain the input value
                 case ValNull => ValNull           // the expression can't be compared to the input value
+                case erroredVal: ErroredVal => erroredVal // propagate the errored value, from checkEquality
                 case _       => ValBoolean(false) // the expression is not the input value
               }
           }
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -40,6 +40,20 @@ import java.time.{Duration, Period}
 class FeelInterpreter {
 
   def eval(expression: Exp)(implicit context: EvalContext): Val = {
+    if (Thread.interrupted()) {
+      throw new InterruptedException()
+    }
+
+    val result = evalInternal(expression)
+    result match {
+      case ErroredVal(value, _) =>
+        value
+      case _ => result
+    }
+  }
+
+
+  private def evalInternal(expression: Exp)(implicit context: EvalContext): Val = {
     // Check if the current thread was interrupted, otherwise long-running evaluations can not be interrupted and fully block the thread
     if (Thread.interrupted()) {
       throw new InterruptedException()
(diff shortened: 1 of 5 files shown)


saig0:
@spearsandshields thank you for your contribution. :tada: 

I'll review your PR this week. :eyes:

spearsandshields:
@saig0 , thanks. 
I just pushed out a fix to ensure the tests are passing. 
Let me know if there is anything you need from me :)

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -715,23 +739,37 @@ class FeelInterpreter {
...
+          expression match {
+            case comparison: Comparison =>
+              if (comparison.x == ConstInputValue || comparison.y == ConstInputValue) {
+                x match {
+                  case ErroredVal(value, _)          => return value
+                  case _ => // Continue
+                }
+              }
+            case _ => // Continue if not a Comparison or ConstInputValue not found
+          }
```
saig0, line 756: :x: This modification solves the case for the comparison (`? > null`). However, it doesn't solve the issue in general.

The following expressions still return `true` instead of `null`:
```
// for ? = null
? > 20 and ? < 40
// output: true 

? > 20 or ? < 40
// output: true

?
// output: true
```

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -40,6 +40,20 @@ import java.time.{Duration, Period}
...
+    if (Thread.interrupted()) {
+      throw new InterruptedException()
+    }
+
+    val result = evalInternal(expression)
+    result match {
+      case ErroredVal(value, _) =>
+        value
+      case _ => result
+    }
```
saig0, line 52: :wrench: Having two different evaluation methods with the same signature makes the code harder to understand and could lead to issues.

src/main/scala/org/camunda/feel/syntaxtree/Val.scala
```diff
@@ -287,6 +288,10 @@ case class ValError(error: String) extends Val {
   override def toString: String = s"error(\"$error\")"
 }
 
+case class ErroredVal(value: Val, errorType: EvaluationFailureType)  extends Val {
+  override def toString: String = value.toString
+}
```
saig0, line 293: :x: The new type is only used in combination with `ValNull`. It seems like the failure type should be an additional property of `ValNull`. And, it is used only within the interpreter.

saig0 (changes requested):
@spearsandshields the solution solves the concrete issue. :+1: However, I don't see a solution for the general issue with the unary-tests (see the examples in the comment).

I would aim for a general solution that reflects the description of the DMN specification (i.e. an expression with `?` must evaluate to `true`). 

At the moment, I can't share a concrete design. But, this is not the way, sorry.

saig0:
Close this PR in favor of #887.

@spearsandshields thank you for your attempt. :bow:

spearsandshields:
Thanks for drafting this more thorough fix. 
I see the backport prs etc, but unsure what we need to do to receive these updates. Is it merged into the latest version of the camunda engine yet?

saig0:
> I see the backport prs etc, but unsure what we need to do to receive these updates. Is it merged into the latest version of the camunda engine yet?

@spearsandshields I will build new patch releases of the FEEL engine soon. So, the fix will be included in the upcoming Camunda patch releases. :building_construction:

spearsandshields:
@saig0 , thanks and this is huge!

Is there a rough timeline where I can expect the patch release?

saig0:
@spearsandshields you can expect the next patch release in the next few weeks. :construction:

PR #887, opened 2024-08-20
fix: Unary-test behavior with special input variable `?`

### LINKED ISSUE #864: [Bug] Incorrect/inconsistent unary test evaluation when nulls are involved.

saig0:
## Description

Fix the behavior of unary-test expressions if the special input variable `?` is used to access the input value. From the following definition, the current behavior is incorrect in case d).

image

**Key idea:**
- Distinguish the case d) from the other cases  
- Limit the access to the special input variable `?` to unary-tests
- Introduce a new error value type to detect the access of `?` and interrupt the evaluation

**Reviewer hint:** 
The first 3 commits are the most relevant. The rest contains a lot of refactoring and may be harder to compare with the original code.

## Related issues

closes #864

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -34,6 +34,7 @@ import org.camunda.feel.{
 }
 
 import java.time.{Duration, Period}
+import scala.reflect.ClassTag
 
 /** @author
   *   Philipp Ossler
@@ -79,15 +80,15 @@ class FeelInterpreter {
 
       // simple unary tests
       case InputEqualTo(x)                              =>
-        withVal(input, i => checkEquality(i, eval(x), _ == _, ValBoolean))
+        withVal(getImplicitInputValue, i => checkEquality(i, eval(x), _ == _, ValBoolean))
       case InputLessThan(x)                             =>
-        withVal(input, i => dualOp(i, eval(x), _ < _, ValBoolean))
+        withVal(getImplicitInputValue, i => dualOp(i, eval(x), _ < _, ValBoolean))
       case InputLessOrEqual(x)                          =>
-        withVal(input, i => dualOp(i, eval(x), _ <= _, ValBoolean))
+        withVal(getImplicitInputValue, i => dualOp(i, eval(x), _ <= _, ValBoolean))
       case InputGreaterThan(x)                          =>
-        withVal(input, i => dualOp(i, eval(x), _ > _, ValBoolean))
+        withVal(getImplicitInputValue, i => dualOp(i, eval(x), _ > _, ValBoolean))
       case InputGreaterOrEqual(x)                       =>
-        withVal(input, i => dualOp(i, eval(x), _ >= _, ValBoolean))
+        withVal(getImplicitInputValue, i => dualOp(i, eval(x), _ >= _, ValBoolean))
       case InputInRange(range @ ConstRange(start, end)) =>
         unaryOpDual(eval(start.value), eval(end.value), isInRange(range), ValBoolean)
 
@@ -100,16 +101,17 @@ class FeelInterpreter {
       case Division(x, y)        => withValOrNull(divOp(eval(x), eval(y)))
       case Exponentiation(x, y)  =>
         withValOrNull(
-          dualNumericOp(
+          withNumbers(
             eval(x),
             eval(y),
-            (x, y) =>
-              if (y.isWhole) {
+            (x, y) => {
+              val result: Number = if (y.isWhole) {
                 x.pow(y.toInt)
               } else {
                 math.pow(x.toDouble, y.toDouble)
-              },
-            ValNumber
+              }
+              ValNumber(result)
+            }
           )
         )
       case ArithmeticNegation(x) =>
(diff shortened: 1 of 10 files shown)


src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
```diff
@@ -715,25 +724,37 @@ class FeelInterpreter {
...
-              case ValBoolean(true)             => ValBoolean(true)  // the expression is the input value
-              case _ if x == ValBoolean(false)  => ValBoolean(false) // the expression is false
+      inputValue =>
+        eval(expression) match {
+          case _: ValFatalError                       =>
+            eval(expression)(context.add(INPUT_VALUE_SYMBOL -> inputValue)) match {
+              case ValBoolean(true)  => ValBoolean(true)
+              case ValBoolean(false) => ValBoolean(false)
+              case _                 => ValNull
+            }
```
remcowesterhoud, line 701: ❓ I just want to verify I understand this correct:

1. We try to evaluate the expression with the original input value (null). Somehow this results in a fatal error.
2. We evaluate the expression again setting `?` to the input value (null) in the context. This will evaluate properly to:
    1. True if the expression evaluates to true (eg: `? = null`)
    2. False if the expression evaluates to false (eg `? != null`)
    3. Null if the expression doesn't return a boolean (eg `odd(?)`)


Before the evaluation would result in True because:

1. The fatal error didn't stop the pattern matching early. As a result we'd check equality: `checkEquality(i, x, _ == _, ValBoolean)`
2. `i` and `x` are both null, so they're equal resulting in True.

The entire concept of `?` didn't seem to exist before?

remcowesterhoud (approved):
Thanks @saig0. I think I understand what's happening but I raised 1 question to be sure. Please have a look 🙂

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
```diff
@@ -511,37 +511,112 @@ class InterpreterUnaryTest extends AnyFlatSpec with Matchers with FeelIntegratio
...
+  it should "return true if it evaluates to true when null is assigned to the special variable '?'" in {
+
+    evalUnaryTests(null, "? = null") should be(ValBoolean(true))
+    evalUnaryTests(null, "? = null or odd(?)") should be(ValBoolean(true))
+  }
+
+  it should "return false if it evaluates to false when null is assigned to the special variable '?'" in {
+
+    evalUnaryTests(null, "? != null") should be(ValBoolean(false))
+    evalUnaryTests(null, "? != null and odd(?)") should be(ValBoolean(false))
```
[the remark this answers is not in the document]
saig0: Good point. :+1:

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterUnaryTest.scala
```diff
@@ -511,37 +511,112 @@ class InterpreterUnaryTest extends AnyFlatSpec with Matchers with FeelIntegratio
...
+
+    evalUnaryTests(5, "null") should be(ValBoolean(false))
+    evalUnaryTests(5, "2 + not_existing") should be(ValBoolean(false))
+    evalUnaryTests(5, "not_existing") should be(ValBoolean(false))
+  }
+
+  it should "return true if it evaluates to true when null is assigned to the special variable '?'" in {
+
+    evalUnaryTests(null, "? = null") should be(ValBoolean(true))
+    evalUnaryTests(null, "? = null or odd(?)") should be(ValBoolean(true))
```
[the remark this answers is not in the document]
saig0: True. :+1:

src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
saig0: Correct. :100: 

 > The entire concept of ? didn't seem to exist before?

Before, `?` was replaced with the input value. There was no check that the expression must return true. It was not strict enough.

remcowesterhoud: > Before, ? was replaced with the input value. There was no check that the expression must return true. It was not strict enough.

Thanks for the extra context! All good to merge imo 😄

PR #889, opened 2024-08-21
fix: Return `null` if a date is invalid

### LINKED ISSUE #867: The function `date()` doesn't return `null` for invalid dates
**Describe the bug**
I can invoke the function `date()` with an invalid date, for example, `"2023-02-29"`. I expected `null` but the function returns the last date of the month.

**To Reproduce**
1. Evaluate the expression `date("2023-02-29")`
2. Verify that the result is a date at `"2023-02-28"`

**Expected behavior**
The function `date()` returns `null` if the date is invalid.

To be consistent, we should also return `null` for date literals, date-time literals, and the `date and time()` function if the date is invalid.

**Environment**
* FEEL engine version: `1.17.7`
* Affects: 
  * Zeebe broker: [0.x] 
  * Support: https://jira.camunda.com/browse/SUPPORT-22372
* OMG: https://issues.omg.org/browse/DMN17-67

saig0:
## Description

Adjust the behavior of date and date-time literals and conversion functions to return `null` if the date is invalid. Previously, it returned the last day of the month. 

For example:
- `2023-02-29` is invalid because 2023 is not a leap year
- `2023-06-31` is invalid because June has only 30 days

## Related issues

closes #867

### DIFF
--- a/src/main/scala/org/camunda/package.scala
+++ b/src/main/scala/org/camunda/package.scala
@@ -18,20 +18,20 @@ package org.camunda
 
 import scala.math.BigDecimal
 import org.slf4j.LoggerFactory
+
 import java.time.temporal.ChronoField.{
+  DAY_OF_MONTH,
   HOUR_OF_DAY,
   MINUTE_OF_HOUR,
-  SECOND_OF_MINUTE,
-  NANO_OF_SECOND,
-  YEAR,
   MONTH_OF_YEAR,
-  DAY_OF_MONTH
+  NANO_OF_SECOND,
+  SECOND_OF_MINUTE,
+  YEAR
 }
-
 import java.util.regex.Pattern
 import org.camunda.feel.syntaxtree.ZonedTime
 
-import java.time.format.{DateTimeFormatterBuilder, SignStyle}
+import java.time.format.{DateTimeFormatterBuilder, ResolverStyle, SignStyle}
 import java.time.{Duration, LocalDate, LocalDateTime, LocalTime, Period, ZonedDateTime}
 
 /** @author
@@ -189,17 +189,20 @@ package object feel {
     .appendValue(MONTH_OF_YEAR, 2)
     .appendLiteral("-")
     .appendValue(DAY_OF_MONTH, 2)
-    .toFormatter();
+    .toFormatter()
+    .withResolverStyle(ResolverStyle.STRICT)
 
   val localDateTimeFormatter = new DateTimeFormatterBuilder()
     .append(dateFormatter)
     .append(timeFormatterWithPrefix)
-    .toFormatter();
+    .toFormatter()
+    .withResolverStyle(ResolverStyle.STRICT)
 
   val dateTimeFormatter = new DateTimeFormatterBuilder()
     .append(dateFormatter)
     .append(timeFormatterWithPrefix)
     .append(offsetFormatter)
-    .toFormatter();
+    .toFormatter()
+    .withResolverStyle(ResolverStyle.STRICT)
 
 }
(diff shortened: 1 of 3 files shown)


src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
```diff
@@ -63,6 +63,18 @@ class BuiltinConversionFunctionsTest
     )
   }
 
+  it should "return null if the date is not valid (leap year)" in {
+
+    evaluateExpression(""" date(x) """, Map("x" -> "2023-02-29")) should returnNull()
```
korthout, line 68: 🔧 2023-02-29 is invalid because it is **not a leap year**, but the test description imply otherwise.

```suggestion
  it should "return null if the date is not valid (not a leap year)" in {

    evaluateExpression(""" date(x) """, Map("x" -> "2023-02-29")) should returnNull()
```

src/test/scala/org/camunda/feel/impl/builtin/BuiltinConversionFunctionsTest.scala
```diff
@@ -63,6 +63,18 @@ class BuiltinConversionFunctionsTest
...
+    evaluateExpression(""" date(2023, 2, 29) """) should returnNull()
+  }
+
+  it should "return null if the date is not valid (month with 31 days)" in {
+
+    evaluateExpression(""" date(x) """, Map("x" -> "2023-06-31")) should returnNull()
```
korthout, line 74: 🔧 2023-06-31 is a month without 31 days, but the test seems to imply otherwise:

```suggestion
  it should "return null if the date is not valid (month without 31 days)" in {

    evaluateExpression(""" date(x) """, Map("x" -> "2023-06-31")) should returnNull()
```

src/main/scala/org/camunda/package.scala
```diff
@@ -189,17 +189,20 @@ package object feel {
     .appendValue(MONTH_OF_YEAR, 2)
     .appendLiteral("-")
     .appendValue(DAY_OF_MONTH, 2)
-    .toFormatter();
+    .toFormatter()
+    .withResolverStyle(ResolverStyle.STRICT)
```
korthout, line 193: 👍 Interesting solution

💭 Sadly [the JavaDoc](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/time/format/ResolverStyle.html#STRICT) is not explicit with examples. Still, your tests cover the interesting edge cases that we care about so that's alright.

src/main/scala/org/camunda/package.scala
```diff
@@ -18,20 +18,20 @@ package org.camunda
...
   HOUR_OF_DAY,
   MINUTE_OF_HOUR,
-  SECOND_OF_MINUTE,
-  NANO_OF_SECOND,
-  YEAR,
   MONTH_OF_YEAR,
-  DAY_OF_MONTH
+  NANO_OF_SECOND,
+  SECOND_OF_MINUTE,
+  YEAR
```
korthout, line 29: 💭 Alphabetical ordering 🤷

src/test/scala/org/camunda/feel/impl/interpreter/InterpreterLiteralExpressionTest.scala
```diff
@@ -16,285 +16,216 @@
...
   }
 
   it should "be a context (identifier as key)" in {
 
-    eval("{ a : 1 }")
-      .asInstanceOf[ValContext]
-      .context
-      .variableProvider
-      .getVariables should be(Map("a" -> ValNumber(1)))
+    evaluateExpression("{ a : 1 }") should returnResult(Map("a" -> 1))
```
korthout, line 64: 😍

korthout (approved):
Great stuff @saig0 🗓️ 

Just a few minor comments

👍 LGTM

src/main/scala/org/camunda/package.scala
saig0: I found the trick with the resolver style when I checked the default behavior of `LocalDate.parse()`, which fails if the date is invalid.

Issue #65, opened 2019-06-26
Simplify implementation


saig0:
In order to make it easier for developers to contribute, we should simplify the implementation. For example, get rid of implicit conversions.

PR #892, opened 2024-08-22
fix: number() to return null if the given string is invalid

### LINKED ISSUE #731: number() should return null instead of failing the evaluation

saig0:
## Description

Fix the `number()` function to return `null` if the given string is invalid. Before the evaluation failed with an exception. 

## Related issues

closes #731

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
@@ -170,16 +170,25 @@ class ConversionBuiltinFunctions(valueMapper: ValueMapper) {
     builtinFunction(
       params = List("from"),
       invoke = { case List(ValString(from)) =>
-        ValNumber(from)
+        parseNumber(from)
       }
     )
+
+  private def parseNumber(from: String): Val = {
+    Try(
+      ValNumber(from)
+    ).getOrElse(
+      ValError(s"Can't parse '$from' as a number")
+    )
+  }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
```diff
@@ -170,16 +170,25 @@ class ConversionBuiltinFunctions(valueMapper: ValueMapper) {
     builtinFunction(
       params = List("from"),
       invoke = { case List(ValString(from)) =>
-        ValNumber(from)
+        parseNumber(from)
       }
     )
 
+  private def parseNumber(from: String): Val = {
```
nicpuppa, line 177: ❓ Is it possible to move this parsing as built in behaviour for `ValNumber` ?

nicpuppa (approved):
LGTM, I have just a question 🚀

src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
saig0: Yes, this would be possible. As part of #65, I plan to remove the implicit conventions from the `package` object and add construction methods to the value types. 

However, this will be part of a bigger refactoring. :telescope:

PR #893, opened 2024-08-22
fix: substring() with exceeding length

### LINKED ISSUE #801: The evaluation of `substring()` fails if the length is greater than the given string
**Describe the bug**
If I invoke the [substring()](https://docs.camunda.io/docs/8.3/components/modeler/feel/builtin-functions/feel-built-in-functions-string/#substringstring-start-position-length) function with a length that is greater than the given string then the evaluation fails and reports a Java `StringIndexOutOfBoundsException`. 

**To Reproduce**
1. Evaluate the expression `substring("abc", 1, 4)`
2. Verify that the evaluation fails with the message

```
failed to evaluate expression 'substring("abc", 1, 4)' :
 java.lang.StringIndexOutOfBoundsException: begin 0, end 4, length 3
```

**Expected behavior**

The function invocation should return the whole string if the length is greater than the string. 

```
substring("abc", 1, 4)    // --> "abc"
```

---

However, according to the DMN 1.5 specification, the function invocation should return `null`.

Screenshot from 2024-02-01 15-26-04

However, returning `null` is not the most (business) user-friendly behavior. It could lead to complicated expressions, for example:

```
if string length(message) < 10 then message else substring(message, 1, 10)
```

**Environment**
* FEEL engine version: `1.17.4`
* Affects: 
  * Zeebe broker: `8.4.0`

saig0:
## Description

Fix the behavior of the function `substring()`. If the given length is greater than the string, return the remaining characters from the start index instead of throwing an exception.  

## Related issues

closes #801

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala
@@ -57,11 +57,11 @@ object StringBuiltinFunctions {
   private def substringFunction3 = builtinFunction(
     params = List("string", "start position", "length"),
     invoke = { case List(ValString(string), ValNumber(start), ValNumber(length)) =>
+      val startIndex = stringIndex(string, start.intValue)
+      val endIndex   = Math.min(startIndex + length.intValue, string.length)
+
       ValString(
-        string.substring(
-          stringIndex(string, start.intValue),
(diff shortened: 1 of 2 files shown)


megglos (approved):
LGTM! The previous used `eval` is just superseded by the new `evaluateExpression` matcher used and you gradually migrate?

PR #895, opened 2024-08-23
fix: List operation returns null if the value is not a list

### LINKED ISSUE #809: List filter expression throws error if list is null
**Describe the bug**

If I try to filter a list that does not exist (is null), the expression will throw an error like:
```
failed to evaluate expression 'partner[item.nummer = "1"]': Expected list but found 'null'
```
Also, there are warnings that represent the same:
```
* [NO_VARIABLE_FOUND] No variable found with name 'partner'
* [INVALID_TYPE] Expected list but found 'null'
```
**To Reproduce**

1. Use https://camunda.github.io/feel-scala/docs/playground/
2. Set expression to: `partner[item.nummer = "1"]`
3. Hit evaluate

**Expected behavior**

The result should be null, the warnings are fine

**Environment**
* FEEL engine version: 1.17
* Affects: 
  * Camunda Connector Runtime 8.4.3

saig0:
## Description

Adjust the behavior to return `null` if a list operation (filter, access, some, every) is applied to a value that is not a list.

## Related issues

closes #809

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
+++ b/src/main/scala/org/camunda/feel/impl/interpreter/FeelInterpreter.scala
@@ -169,48 +169,59 @@ class FeelInterpreter {
 
       // list
       case SomeItem(iterators, condition)  =>
-        withCartesianProduct(
-          iterators,
-          p =>
-            atLeastOneValue(p.map(vars => () => eval(condition)(context.addAll(vars))), ValBoolean)
+        withValOrNull(
+          withCartesianProduct(
+            iterators,
+            p =>
+              atLeastOneValue(
+                p.map(vars => () => eval(condition)(context.addAll(vars))),
+                ValBoolean
+              )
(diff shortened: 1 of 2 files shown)


korthout (approved):
LGTM 👍 

👍 Simple changes, easy to review

💭 Except for 65cf32d, which is not really possible to review by reading the changes. I had a look at the resulting tests, and they look good, but I may have missed the removal of a specific useful 
case.

korthout:
>Backport failed for 1.16, because it was unable to cherry-pick the commit(s).

💡 @saig0 https://github.com/korthout/backport-action?tab=readme-ov-file#conflict_resolution

saig0:
Actually, backporting the fix to version `1.16` doesn't make sense. Only since version `1.17`, the FEEL engine is null-friendly. In older versions, the evaluation failed, for example, if a value had a wrong type. So, let's keep the current behavior for version `1.16`.

PR #899, opened 2024-08-23
[Backport 1.16] fix: number() to return null if the given string is invalid

### LINKED ISSUE #731: number() should return null instead of failing the evaluation

saig0:
## Description

Backport of #892 for `1.16`.

Additionally to the changes in the original, this PR contains the migration of the test class to the new test style. This migration was necessary to cherry-pick the commit with the new test cases.

## Related issues

closes #731


nicpuppa (approved):
❤️

PR #902, opened 2024-08-23
fix: Function `context put()` works for more than two keys

### LINKED ISSUE #835: context put does not work correctly for path longer than 2
**Describe the bug**

Function `context put` does not work correctly for path longer than 2. For path with length 1 or 2 it works correctly - it merge _object_ with _value_ on given _path_. For _path_ longer than 2 it returns object with **last** two keys from _path_ without previous data from given _object_.

**To Reproduce**

context:

```json
{"x": {"z": "old"} }
```

expression:

```feel
context put(x, ["a", "b", "c"], "new")
```

result:

```json
{"b": {"c": "new"}}
```

=> Key `a` is ignored and previous data `{"z": "old"}` are also dropped.

**Expected behavior**

```json
{"z": "old", "a": {"b": {"c": "new"}}}
```

**Environment**
* FEEL engine version: [1.17.3] on https://camunda.github.io/feel-scala/docs/playground/

saig0:
## Description

Fix the function `context put()` if more than two keys are passed as arguments.

We had an issue in the recursion and forgot to pass the context updater function with us. As a result, this worked only for one or two keys. :loop: 

## Related issues

closes #835

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ContextBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ContextBuiltinFunctions.scala
@@ -150,10 +150,12 @@ class ContextBuiltinFunctions(valueMapper: ValueMapper) {
           parentContextUpdater =
             // pass a lambda to update this context with the modified nested context
             nestedContext =>
-              contextPut(
-                contextValue = contextValue,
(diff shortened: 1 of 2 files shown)


mustafadagher (approved):
LGTM! 🚀

PR #906, opened 2024-08-26
[Backport 1.16] fix: Unary-test behavior with special input variable `?`

### LINKED ISSUE #864: [Bug] Incorrect/inconsistent unary test evaluation when nulls are involved.

saig0:
## Description

Backport of #887 for `1.16`.

Most changes are the same as in the original. However, I had to adjust the code to align with the error behavior in version `1.16.`. Instead of returning `null` (e.g. if a variable doesn't exist), the evaluation fails or returns `false`. The last commit highlights the different behavior for unary-tests compared to version `1.17+`. 

## Related issues

closes #864

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -328,6 +328,17 @@
               <className>org/camunda/feel/syntaxtree/Range</className>
               <differenceType>8001</differenceType>
             </ignored>
+            <!-- ignore auto-generated Scala functions -->
+            <ignored>
+              <className>org/camunda/feel/syntaxtree/**</className>
+              <differenceType>7002</differenceType>
+              <method>scala.Function1 andThen(scala.Function1)</method>
+            </ignored>
+            <ignored>
+              <className>org/camunda/feel/syntaxtree/**</className>
+              <differenceType>7002</differenceType>
+              <method>scala.Function1 compose(scala.Function1)</method>
+            </ignored>
           </ignored>
         </configuration>
       </plugin>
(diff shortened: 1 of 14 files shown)


remcowesterhoud (approved):
LGTM! 🚀

PR #914, opened 2024-08-28
fix: Parse expression with NBSP character

### LINKED ISSUE #840: Fail to parse expression with NBSP character
**Describe the bug**
The engine fails to parse an expression with a [NBSP](https://en.wikipedia.org/wiki/Non-breaking_space) (Non-breaking Space) character.  

**To Reproduce**
1. Parse an expression with a NBSP character (`<0xa0>`)
2. Verify that the parsing failed

<pre>
"The engine" should "evaluate expression wirth NBSP character" in {
    evaluateExpression("1 +   2") should returnResult(3)
  }
</pre>

**Expected behavior**
The engine can parse expressions with NBSP characters. It should be interpreted as whitespace. 

image

**Environment**
* FEEL engine version: `1.17.3`
* Affects: 
  * Zeebe broker: [0.x] 
  * Support: https://jira.camunda.com/browse/SUPPORT-21217

saig0:
## Description

Adjust the parser's whitespace handling to ignore FEEL-specific whitespace characters, such as NBSP (i.e. Non-breaking space / `\u00A0`),  as defined in the DMN specification.

The `JavaWhitespace` class worked very well because it also supported Java-like comments, as required by the DMN specification. Instead of adding additional whitespace rules for the individual parsers, extend the `JavaWhitespace` class by making a copy of it. :see_no_evil: This should be less error-prone, easier to maintain, and more efficient. :angel: 

## Related issues

closes #840

### DIFF
--- a/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
@@ -21,6 +21,7 @@ import org.camunda.feel.impl.{EvaluationResultMatchers, FeelEngineTest}
 import org.camunda.feel.syntaxtree._
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
+import org.scalatest.prop.TableDrivenPropertyChecks
 
 /** @author
   *   Philipp Ossler
@@ -29,7 +30,8 @@ class InterpreterExpressionTest
     extends AnyFlatSpec
     with Matchers
     with FeelEngineTest
-    with EvaluationResultMatchers {
+    with EvaluationResultMatchers
+    with TableDrivenPropertyChecks {
 
   "An expression" should "be an if-then-else (with parentheses)" in {
     val exp = """ if (x < 5) then "low" else "high" """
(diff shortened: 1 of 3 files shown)


src/test/scala/org/camunda/feel/impl/interpreter/InterpreterExpressionTest.scala
```diff
@@ -364,4 +364,70 @@ class InterpreterExpressionTest
...
+  List(
+    '\u0009', '\u0020', '\u0085', '\u00A0', '\u1680', '\u180E', '\u2000', '\u2001', '\u2002',
+    '\u2003', '\u2004', '\u2005', '\u2006', '\u2007', '\u2008', '\u2009', '\u200A', '\u200B',
+    '\u2028', '\u2029', '\u202F', '\u205F', '\u3000', '\uFEFF', '\u000A', '\u000B', '\u000C',
+    '\u000D'
+  ).foreach { character =>
+    it should f"be parsed (\\u${character.toInt}%04X)" in {
+      evaluateExpression(s"${character}1") should returnResult(1)
+    }
+  }
```
remcowesterhoud, line 431: 🔧 Have you considered putting this in a[ parameterized test](https://www.baeldung.com/scala/scalatest-parameterized-tests#testing-with-parameterized-tests) and running all the scenarios in this class for all whitespace characters?

remcowesterhoud (approved):
Thanks @saig0 🚀 Please have a look a my comment, otherwise all good!

PR #927, opened 2024-09-19
feat: add function `is blank()` to check if a string is blank

skayliu:
## Description

feat: add function `is blank()` to check if a string is blank

## Related issues

related to [https://github.com/camunda/camunda/issues/22380](https://github.com/camunda/camunda/issues/22380)

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/BooleanBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/BooleanBuiltinFunctions.scala
@@ -25,8 +25,7 @@ object BooleanBuiltinFunctions {
     "not"         -> List(notFunction),
     "is defined"  -> List(isDefinedFunction),
     "get or else" -> List(getOrElse),
-    "assert"      -> List(assertFunction, assertFunction2),
-    "is blank"  -> List(isBlankFunction)
+    "assert"      -> List(assertFunction, assertFunction2)
   )
 
   private def notFunction =
@@ -68,13 +67,6 @@ object BooleanBuiltinFunctions {
     invoke = {
       case List(value, ValBoolean(true), _) => value
       case List(_, _, ValString(cause))     => ValError(cause)
-    }
-  )
-
-  private def isBlankFunction = builtinFunction(
-    params = List("string"),
-    invoke = {case List(ValString(string)) =>
-      ValBoolean(string.isBlank)
     }
   )
 
--- a/src/main/scala/org/camunda/feel/impl/builtin/BooleanBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/BooleanBuiltinFunctions.scala
@@ -25,7 +25,8 @@ object BooleanBuiltinFunctions {
     "not"         -> List(notFunction),
     "is defined"  -> List(isDefinedFunction),
     "get or else" -> List(getOrElse),
-    "assert"      -> List(assertFunction, assertFunction2)
+    "assert"      -> List(assertFunction, assertFunction2),
+    "is blank"  -> List(isBlankFunction)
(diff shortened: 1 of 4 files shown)


src/main/scala/org/camunda/feel/impl/builtin/BooleanBuiltinFunctions.scala
```diff
@@ -70,4 +71,11 @@ object BooleanBuiltinFunctions {
     }
   )
 
+  private def isBlankFunction = builtinFunction(
+    params = List("string"),
+    invoke = {case List(ValString(string)) =>
+      ValBoolean(string.isBlank)
+    }
+  )
```
saig0, line 79: :x: Since `is blank()` is a function for a string argument, we should add the function to [StringBuiltinFunctions](https://github.com/camunda/feel-scala/blob/main/src/main/scala/org/camunda/feel/impl/builtin/StringBuiltinFunctions.scala). 

`BooleanBuiltinFunctions` is for functions with a boolean argument.

src/test/scala/org/camunda/feel/impl/builtin/BuiltinFunctionsTest.scala
```diff
@@ -243,4 +243,27 @@ class BuiltinFunctionsTest
       "Assertion failure on evaluate the expression 'list contains(assert(my_list, my_list != null, \"The condition is not true\"), 2)': The condition is not true"
     )
   }
+
+  "A is blank() function" should "return Boolean" in {
```
saig0, line 247: :x: Same as the implementation, we should add the test cases to [BuiltinStringFunctionsTest](https://github.com/camunda/feel-scala/blob/main/src/test/scala/org/camunda/feel/impl/builtin/BuiltinStringFunctionsTest.scala).

src/test/scala/org/camunda/feel/impl/builtin/BuiltinFunctionsTest.scala
```diff
@@ -243,4 +243,27 @@ class BuiltinFunctionsTest
...
+      expression = """ is blank("hello world") """
+    ) should returnResult(false)
+
+    evaluateExpression(
+      expression = """ is blank(" hello world ") """
+    ) should returnResult(false)
+
+    evaluateExpression(
+      expression = """ is blank("\t\n\r\f") """
+    ) should returnResult(true)
```
saig0, line 266: :wrench: We could split the test cases into:

- `should "return true if the string contains only whitespace"`
- `should "return false if the string contains non-whitespace characters"`

src/test/scala/org/camunda/feel/impl/builtin/BuiltinFunctionsTest.scala
```diff
@@ -243,4 +243,27 @@ class BuiltinFunctionsTest
       "Assertion failure on evaluate the expression 'list contains(assert(my_list, my_list != null, \"The condition is not true\"), 2)': The condition is not true"
     )
   }
+
+  "A is blank() function" should "return Boolean" in {
+    evaluateExpression(
+      expression = """ is blank("") """
+    ) should returnResult(true)
```
saig0, line 250: :x: Add one more test case that verifies the invocation of the function with named parameters:

```
evaluateExpression(
      expression = """ is blank(string: "") """
    ) should returnResult(true)
```

This kind of test ensures that we don't change the parameter name accidentally.

saig0 (changes requested):

The changes look good. I have only a few minor comments. Please have a look. :cookie:

skayliu:
Hi, @saig0 , It's done, please check again.

saig0 (approved):
@skayliu looks good. Thank you for the fast response. :rocket:

PR #928, opened 2024-09-19
refactor: extend `is empty()` to check if a string is empty

skayliu:
## Description

Refactor: extend `is empty()` to check if a string is empty

## Related issues

closes #

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/ListBuiltinFunctions.scala
@@ -513,8 +513,9 @@ class ListBuiltinFunctions(private val valueMapper: ValueMapper) {
   private def emptyFunction =
     builtinFunction(
       params = List("list"),
-      invoke = { case List(ValList(list)) =>
-        ValBoolean(list.isEmpty)
+      invoke = {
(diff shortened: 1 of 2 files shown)


skayliu:
> @skayliu thank you for taking the time to contribute. 🙇
> 
> First, I would like to clarify if we need the function `is empty(string)` for a string value. I would assume that we would cover the use cases with the new function `is blank(string)` ([ref](https://github.com/camunda/feel-scala/pull/927)).
> 
> Why do you see a need for this function? Please describe your use case.

Because `is empty(string)` is different from `is blank(string)` , so i think we need it.

saig0:
> Because is empty(string) is different from is blank(string) , so i think we need it.

Yes, the behavior differs in the following cases:

```
is blank("")     // true
is empty("")     // true 

is blank(" ")    // true
is empty(" ")    // false 
```

However, I challenge that we need both functions. Please come up with a case where you need `is empty()` and can't use `is blank()`?

---

Reasoning: 
- Adding a new function increases the maintenance of our tools and the complexity of the language. 
- We aim to add all functions to the DMN standard. The OMG needs to approve the proposal. So, we should provide good arguments for it.

skayliu:
@saig0 ,Let's close this PR for less good reason.

skayliu:
@saig0, Let's close this PR for less good reason.

PR #930, opened 2024-09-24
fix: Handle escape and regex characters

### LINKED ISSUE #879: Incorrect escaping in strings
**Describe the bug**
Incorrect escping of backslash

**To Reproduce**
1. Open playground: https://camunda.github.io/feel-scala/docs/playground/
2. Use expression `x = "\\"`
3. Use  context `{ "x": "\\" }`

**Observed behavior**
image

**Expected behavior**
Output should be **true**

**Notes**
If expression is `x`, then in output there is string with 2 slashes instead of one
If expression is `"\\"` then currently output is string with 4 slashes, which is very srange

**Environment**
* FEEL engine version: [1.17]

### LINKED ISSUE #883: FEEL engine replaces regex expressions in output
**Describe the bug**

**To Reproduce**
Using Camunda Platform 7:
1. Deploy a simple DMN with a regex string (e.g., test\s\r\ntest in the output column.
2. Execute an instance (e.g., through BPMN or REST API) triggering the row with the regex output.
3. Check the output in Cockpit. The displayed regex string was changed. The escaped characters are transformed.

<details><summary>Example DMN</summary>

```
<?xml version="1.0" encoding="UTF-8"?>
<definitions xmlns="https://www.omg.org/spec/DMN/20191111/MODEL/" xmlns:dmndi="https://www.omg.org/spec/DMN/20191111/DMNDI/" xmlns:dc="http://www.omg.org/spec/DMN/20180521/DC/" xmlns:modeler="http://camunda.org/schema/modeler/1.0" xmlns:camunda="http://camunda.org/schema/1.0/dmn" id="Definitions_0g2ba0p" name="DRD" namespace="http://camunda.org/schema/1.0/dmn" exporter="Camunda Modeler" exporterVersion="5.21.0" modeler:executionPlatform="Camunda Platform" modeler:executionPlatformVersion="7.20.0">
  <decision id="regex" name="regex" camunda:historyTimeToLive="90">
    <decisionTable id="DecisionTable_1ac01ar">
      <input id="Input_1" label="Student Class" camunda:inputVariable="student_class">
        <inputExpression id="InputExpression_1" typeRef="integer">
          <text>student_class</text>
        </inputExpression>
      </input>
      <output id="Output_1" label="Regex" name="regex" typeRef="string" />
      <rule id="DecisionRule_0pa34gk">
        <inputEntry id="UnaryTests_1o9hf59">
          <text>1</text>
        </inputEntry>
        <outputEntry id="LiteralExpression_0oxhb51">
          <text>"\s\r\n"</text>
        </outputEntry>
      </rule>
    </decisionTable>
  </decision>
  <dmndi:DMNDI>
    <dmndi:DMNDiagram>
      <dmndi:DMNShape dmnElementRef="regex">
        <dc:Bounds height="80" width="180" x="160" y="100" />
      </dmndi:DMNShape>
    </dmndi:DMNDiagram>
  </dmndi:DMNDI>
</definitions>

```

</details> 

**Expected behavior**
The output of the triggered DMN row is not changed. The escaped characters are not transformed. The original regex string is returned as output.

**Solution idea**
* FEEL engine can handle escape and regex characters properly. 
* `\s` should not be replaced with a whitespace but handled the same way as `\n` and `\r`.  

**Environment**
* FEEL engine version: [1.17.5], but newer versions should have the same behavior
* Affects: 
  * Camunda Automation Platform 7: [7.22.0] https://github.com/camunda/camunda-bpm-platform/issues/4530
* Support case: https://jira.camunda.com/browse/SUPPORT-23111

saig0:
## Description

Correct the handling of escape sequences in string literals.

Don't replace escape sequences in regex expressions, for example, \r or \n. In the parser, these sequences start with \\. Same for \s, don't replace it with a whitespace, since this is also a part of a regex.

Handle \\ to avoid that the sequence is escaped and returned as \\\\.

## Related issues

closes #883
closes #879


saig0:
@mboskamp I fixed the issue with `\s` character. I did an integration test in C8 and it looks good from my perspective. :white_check_mark: 

However, it is important to note that `"\s"` is not a valid string literal because `\s` is **not an escape character**. Instead, it should be `"\\s"`. Same for `\n` and `\r`, if both characters are used for a regex then they should be escaped as `\\n\\r`. As a result, the string in the example DMN should be `"\\s\\r\\n"` instead of `"\s\r\n"`.

mboskamp (approved):
👍 Looks good. I did not test this in C7 and did not fully understand the fix. However, the test cases look good.

PR #959, opened 2024-12-18
fix: A unary-test expression with ? returns true/false

saig0:
## Description

The engine restricts the usage of `?` to an unary-test expression. It is implemented by returning a `ValFatalError` when accessing `?`. If the expression is a unary-test, it evaluates the expression again and assigns the input value to `?`.  

Previously, the engine invoked a function without checking the arguments before. As a result, a function with an `Any` parameter (e.g. `list contains(list, any)`) accepted a `ValError`/`ValFatalError` as a regular value and didn't forward the error. 

Fix the issue by adding the validation of the arguments before invoking the function. 

## Related issues

closes https://github.com/camunda/camunda/issues/26054

### DIFF
--- a/src/main/scala/org/camunda/feel/impl/builtin/BuiltinFunction.scala
+++ b/src/main/scala/org/camunda/feel/impl/builtin/BuiltinFunction.scala
@@ -16,7 +16,7 @@
  */
 package org.camunda.feel.impl.builtin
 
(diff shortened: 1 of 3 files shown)


tasso94 (approved):

saig0:
- #964

saig0:
- #965

saig0:
- #966

PR #960, opened 2024-12-23
refactor: lower log level to debug

npepinpe:
## Description

This PR lowers the log level of the initial engine creation log statement. The reasoning here is that almost all of the information provided here is only really useful for debugging, and generally not useful for the common user.

As this is mostly to reduce the noise-to-signal ratio in Zeebe, the alternative would be only showing WARN messages from this library (as we do with other libraries). That would also be acceptable to me, so feel free to reject this if you disagree, and we can go with the alternative :)

## Related issues

closes camunda/camunda#26395


saig0 (approved):
@npepinpe LGTM. :+1:

saig0:
- #961

saig0:
- #962

saig0:
- #963

PR #969, opened 2025-01-16
fix: A unary-test expression returns null instead of false

saig0:
## Description

Fix a regression for unary-test expressions. If the expression evaluates to `false` and the input value is of a different type, the unary-test should return `false` instead of `null`.

## Related issues

closes https://github.com/camunda/camunda/issues/27004


nicpuppa (approved):
LGTM 🥳

saig0:
- #974

saig0:
- #975

PR #993, opened 2025-03-27
eppdot-fix-lazy-value-mapper

eppdot:
## Description

Previously, the builtin context#put function was expecting to
receive a fully mapped context (including all nested variables).
But this can not always be guaranteed as some VariableProviders
can only resolve variables lazily, because of performance/memory
concerns.

For example, in camunda engine we don't want to decode all possible
variables of a process/element-scope at once as most of them might
not be needed by the expression at all.

## Related issues

Closes https://github.com/camunda/camunda/issues/29752

### DIFF
--- a/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
+++ b/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
@@ -16,10 +16,12 @@
  */
 package org.camunda.feel.impl.builtin
 
-import org.camunda.feel.context.{CustomContext, VariableProvider}
+import org.camunda.feel.api.FeelEngineBuilder
+import org.camunda.feel.context.Context
 import org.camunda.feel.impl.interpreter.MyCustomContext
 import org.camunda.feel.impl.{EvaluationResultMatchers, FeelEngineTest}
 import org.camunda.feel.syntaxtree._
+import org.camunda.feel.valuemapper.CustomValueMapper
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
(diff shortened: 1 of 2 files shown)


src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
```diff
@@ -234,6 +236,35 @@ class BuiltinContextFunctionsTest
...
+      })
+      .build()
+
+    lazyEngine.evaluateExpression(
+      """ context put(vars, ["a", "c"], 3) """,
+      Map("vars" -> Map("a" -> Map("b" -> 1, "c" -> 2)))
+    ) should returnResult(
+      Map("a" -> Map("b" -> 1, "c" -> 3))
+    )
+  }
```
saig0, line 266: :wrench: We could simplify the test case and align with [other test cases](https://github.com/camunda/feel-scala/blob/117e335b579b221c6e3bb1c97c88f509c4b4632c/src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala#L119-L133).

```suggestion
  it should "override nested context entry from a custom context" in {
    evaluateExpression(
      expression = """ context put(vars, ["a", "c"], 3) """,
      variables = Map(
        "vars" ->
          ValContext(
            new MyCustomContext(
              Map("a" -> Map("b" -> 1, "c" -> 2))
            )
        )
      )
    ) should returnResult(
      Map("a" -> Map("b" -> 1, "c" -> 3))
    )
  }
```

saig0 (changes requested):
@eppdot thank you for fixing the issue. :rocket: 

The fix looks good. I have a suggestion to improve the test case. Please have a look. :cookie:

src/test/scala/org/camunda/feel/impl/builtin/BuiltinContextFunctionsTest.scala
eppdot: ❤️ like your suggestion. I tried to make it work with a `CustomContext` as well, but have not been successful (I think I had the `CustomContext` at the outermost to provide a `context` to evaluate instead of plain `variables`)

Still, kudos to you about the ValueMapper design. I think the `Optional` return value in combination with the chain of mappers and the `DefaultMapper` is great and nicely extensible 💪

saig0 (approved):
LTGM. :rocket:

saig0:
@eppdot I will plan patch releases for next week. :building_construction:

saig0:
- #996

Issue #602, opened 2023-03-10
Stringify JSON objects


aleksander-dytko:
**Is your feature request related to a problem? Please describe.**
During a conversation with a customer, they mentioned that it would be great to be able to structuralize JSON object with FEEL expression to stringify it. 

**Describe the solution you'd like**
A new built-in function to return a JSON string of a FEEL value, like `JSON.stringify()`.

```
// Function signature
to json(value: Any): String

// Examples
// 1) FEEL context to JSON object
to json({a: 1, b: 2})
// "{\"a\": 1, \"b\": 2}"

// 2) FEEL value to JSON literal
to json(1)
// "1"

to json("a")
// "\"a\""

to json(true)
// "true"

to json(null)
// "null"

// 3) FEEL list to JSON array
to json([1, 2, 3])
// "[1, 2, 3]"

// 4) FEEL date/time/date-time to JSON string
to json(@"2023-06-14")
// "\"2023-06-14\"" 

to json(@"14:55:00")
// "\"14:55:00\"" 

to json(@"2023-06-14T14:55:00")
// "\"2023-06-14T14:55:00\"" 

// 5) FEEL duration to JSON string 
to json(@"P1Y")
// "\"P1Y\"" 

to json(@"PT2H")
// "\"PT2H\"" 

// 6) FEEL function to JSON string 
to json(function (a, b) a + b)
// "\"<function>\"" 

// 7) FEEL range to JSON string 
to json([1..10])
// "\"[1..10]\"" 
```

**Related issues**

* https://github.com/camunda/feel-scala/issues/825
* Camunda Autormation Platform 7: 
* Zeebe broker: 
* OMG issues:
  * https://issues.omg.org/browse/DMN16-68
  * https://issues.omg.org/browse/DMN15-102

_EDIT: I updated the description to focus on producing a JSON string and to propose a new built-in function with examples._

saig0:
@aleksander-dytko please provide more input for this issue.  

What is the actual use case? 
Do you have an example?
In which context is the expression evaluated? BPMN@Zeebe?

What should the function do?

In Zeebe, all variables are stored as JSON objects. In FEEL, we can access these JSON objects naturally as FEEL context objects.

The result of a FEEL expression is always a FEEL type, for example, a FEEL context.

McAlm:
@aleksander-dytko: Another use case would be if a JSON string (coming from some database) is being passed to a DMN decision, e.g. 
```
    {
        "WRAPPER": [
          {
[...]

PR #1038, opened 2025-08-19
feat: add `to json(any): string` conversion function

rnewton-bp3:
## Description
This PR adds a new built-in FEEL function `to Json(object): string` that performs the inverse of the existing `from Json(json):any` function. It allows FEEL expressions to serialize values (contexts, lists, primitives, temporal values, and durations) into JSON strings, similar to `JSON.stringify(any)` function in JavaScript. This is especially useful for integration scenarios where FEEL values need to be passed to systems expecting JSON.

**Implementation details:**

Introduced helper method `convertToJsonValue(Val)` to recursively map FEEL Val types into JSON-compatible Java/Scala values.

Supports all FEEL data types, including temporal values (ValDate, ValTime, ValDateTime, ValLocalTime, ValLocalDateTime) formatted using ISO-8601 standards.

Durations are serialized using their toString representation.

The resulting JSON string is wrapped in a ValString to be fully FEEL-compatible.

Added safety net to return ValError when serialization fails (though in practice all FEEL types are supported).

I could not create a meaningful test that triggers the `ValError(s"Failed to convert object to JSON: ${obj.getClass.getSimpleName}")` case because all FEEL data types are fully mapped in the `convertToJsonValue(Val)` function. Triggering this branch would require artificially passing in unsupported or non-FEEL types, which isn’t a real usage scenario.

Finally, I’m a complete Scala noob, so I appreciate any suggestions for making the code cleaner or more idiomatic.

Thanks to Alex Trukhny (@sashtran) 🙏  for the initial review. 

Related issues
closes https://github.com/camunda/feel-scala/issues/602

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -25,7 +25,7 @@ jobs:
     steps:
       - uses: actions/checkout@v5
       - name: Create backport PRs
-        uses: zeebe-io/backport-action@v3.3.0
+        uses: zeebe-io/backport-action@v3.4.1
         with:
           # Required
           # Token to authenticate requests to GitHub
--- a/pom.xml
+++ b/pom.xml
@@ -15,10 +15,10 @@
   <name>FEEL Scala Engine</name>
 
   <properties>
-    <scala.version>2.13.16</scala.version>
+    <scala.version>2.13.17</scala.version>
     <scala.binary.version>2.13.6</scala.binary.version>
     <version.log4j>2.25.2</version.log4j>
-    <version.uuid>5.1.0</version.uuid>
+    <version.uuid>5.1.1</version.uuid>
     <version.jackson>2.19.2</version.jackson>
 
     <plugin.version.shade>3.6.1</plugin.version.shade>


src/main/scala/org/camunda/feel/impl/builtin/ConversionBuiltinFunctions.scala
```diff
@@ -373,4 +388,31 @@ class ConversionBuiltinFunctions(valueMapper: ValueMapper) {
...
+  private def convertToJsonValue(value: Val): Any = value match {
+    case ValNull       => null
+    case ValString(s)  => s
+    case ValBoolean(b) => b
+    case ValNumber(n)  => n.underlying()
+
+    case ValDate(d)           => d.format(DateTimeFormatter.ISO_LOCAL_DATE)
+    case ValTime(t)           => t.format
+    case ValDateTime(dt)      => dt.format(DateTimeFormatter.ISO_OFFSET_DATE_TIME)
+    case ValLocalTime(t)      => t.format(DateTimeFormatter.ISO_LOCAL_TIME)
```
bschlosser, line 401: Should there be a case for `ValLocalDate` too?

bschlosser: Great, thanks for checking

saig0:
@rnewton-bp3 thank you for your contribution. :tada: I'll review your changes soon. :eyes:

src/main/scala/org/camunda/feel/syntaxtree/ZonedTime.scala
```diff
@@ -164,4 +166,14 @@ object ZonedTime {
...
+  def format(zonedTime: ZonedTime): String = {
+    val base = zonedTime.time.format(timeFormatter)
+
+    val withOffset = base + zonedTime.offset.toString
+
+    zonedTime.zone match {
+      case Some(zoneId) => withOffset + s"[${zoneId.getId}]"
+      case None         => withOffset
+    }
+  }
```
saig0, line 178: :wrench: Let's move this function into the class `ZoneTime` and align it with the `format()` function.

```
def formatToIso: String = {
    val localTime = localTimeFormatter.format(time)
    val withOffset = localTime + offsetFormatter.format(offset)

    if (hasTimeZone) {
      val zoneId = zone.get.getId
      s"$withOffset[$zoneId]"
    } else {
      withOffset
    }
  }
```

saig0 (changes requested):
@rnewton-bp3 looks very good. :+1: 

I have just one minor improvement. Please take a look. :cake:

src/main/scala/org/camunda/feel/syntaxtree/ZonedTime.scala
rnewton-bp3: Updated. Please review again to see if that looks good.

src/main/scala/org/camunda/feel/syntaxtree/ZonedTime.scala
```diff
@@ -114,6 +126,8 @@ object ZonedTime {
     .appendOffsetId()
     .toFormatter()
 
+  private val timeFormatter: DateTimeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss")
+
```
saig0, line 130: 🔧 Remove unused code.


```suggestion
```

saig0 (approved):
@rnewton-bp3 looks great. 🎉 

Thank you for contributing this valuable new feature. 🚀

PR #1083, opened 2025-12-16
perf: improve performance of FeelInterpreter.hasSameType

entangled90:
## Description
Avoids creating intermediate collections and uses `.forall`  to verify that the condition holds for all elements of the list.


saig0 (approved):
👍

PR #1142, opened 2026-03-14
docs: switch maven central badge to shields.io

korthout:
## Description

The current Maven Central badge is no longer functional. It appears the Maven Badges Heroku app is no longer active.

This replaces it with the shields.io one and provides a link to the mvnrepository.com artifact for easy navigation.

## Related issues

no related issue


saig0 (approved):
@korthout thank you. 🫶

PR #1200, opened 2026-09-18
fix: migrate Vault auth from AppRole to JWT/OIDC

kellervater:
## 📌 What

Flips `hashicorp/vault-action` from `method: approle` to `method: jwt` in the `release` job, which had a genuine (non-false-positive) `no-permissions-context` gap: the `mvn release:prepare release:perform` step pushes commits/tags using the ambient checkout token. Adds `permissions: { contents: write, id-token: write }` and pins the touched action to its latest known SHA + semver comment.

## 🤔 Why

camunda/team-infrastructure#1076 is migrating all repos off Vault AppRole to short-lived, per-repository JWT/OIDC auth.

## 🔧 How

Generated by a deterministic transform tool (`camunda/infra-core` `cmd/github/vault-jwt-migration/transform.py`) that only rewrites a call site when the secret references are the plain names and the job's permissions can be safely extended; the `contents: write` grant here was added by hand once the ambient-token usage was confirmed.

## 📚 References

- camunda/team-infrastructure#1076


saig0 (approved):
👍

